PR #2, opened 2022-08-24
0.1.1 - replace Math.random with nanoid

AesSedai:
`Math.random` is not a secure or entropically sufficient source for generating GUIDs, replace with `nanoid` like in the core surrealdb repository.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "surrealdb.js",
-  "version": "0.1.1",
+  "version": "0.1.0",
   "description": "Javascript driver for SurrealDB",
   "main": "index.js",
   "browser": "dist/web/main.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "surrealdb.js",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "description": "Javascript driver for SurrealDB",
   "main": "index.js",
   "browser": "dist/web/main.js",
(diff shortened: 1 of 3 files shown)


tobiemh:
Thank you so much for this @AesSedai ! This is awesome work! I'm presuming this is something that could work / be changed in the [`Deno` library](https://github.com/surrealdb/surrealdb.deno) repository too?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "surrealdb.js",
-  "version": "0.1.0",
+  "version": "0.1.1",
```
tobiemh, line 3: Could you remove this new version (and in the package-lock.json), and I'll bump it separately (with a Git tag too), and update dependencies?

AesSedai:
As far as Deno goes, it would probably work there too? I haven't used Deno before though and I know the package ecosystem is a little different there.

tobiemh:
Hi @AesSedai , I merged this yesterday, but I've just thought of a few issues, as I was implementing it for Deno...

1. The nano library in a browser (as this repository can be run from backend or frontend).
2. The random id, isn't actually used for generating record IDs, but only for generating a rough event id (for receiving notifications from the server). As a result it doesn't actually need to be cryptographically secure, as the id is only for detecting the received response, after sending the websocket notification.

I'm actually thinking of reversing this change now. Would be good to get your thoughts...

tobiemh:
Ignore that @AesSedai this works fine, and I'm keeping it in! Awesome work, thank you!

tobiemh:
This is now in the 0.2.0 release 👏

PR #3, opened 2022-08-28
Converting to typescript

michaelprichardson:
This is to convert the lib to TS. I focused more on adding types to the exported `Surreal` class as this is what the user will see.

I had a suggestion regarding the methods for querying as they can return either a single object or an array of objects. Similarly with the multiple queries. It makes the return types more ambiguous. So for instance with the `create` method, have a `createOne` and `createMultiple` (could come up with better naming). This way it will be possible to know the return type of `T` or `Array<T>`


michaelprichardson:
@tobiemh I realised that converting all the files in one go makes it a bit of a nightmare PR to review. I created a branch with only adding types to the `Surreal` class as that's the user-facing class, here's the [branch](https://github.com/surrealdb/surrealdb.js/compare/main...michaelprichardson:surrealdb.js:convert-to-ts-simplified?expand=1). I tried to keep the changes to a minimum.

Let me know if I should close this PR and open one for that branch.

PR #5, opened 2022-09-06
Start converting js to ts

michaelprichardson:
Added basic typescript config and tried to change as little as possible to get the conversion started. Added types to the arguments and returns in the `Surreal` class. As this is the front-facing class I thought it would be best to start there and it makes the PR small


AesSedai:
FWIW, I've been following your full PR and looks like this is a good starting point for the TS conversion. Excited for the full conversion to merge later.

Trystan-SA:
You should also try to add JSDOC from [the existing documentation](https://surrealdb.com/docs/integration/libraries/nodejs) so that we can have nice autocompletion hints ! 

Here is an example of what I added on my side : 
```ts
	/** Creates a record in the database.
		@param thing REQUIRED - The table name or the specific record id to create.
		@param data OPTIONAL - The document / record data to insert.
		@example // Create a record with a random id
		let person = await db.create('person');
		// Create a record with a specific id
		let record = await db.create('person:tobie', {
			name: 'Tobie',
			settings: {
				active: true,
				marketing: true,
			},
		});
		//This function will run the following query in the database
		CREATE $thing CONTENT $data;
	*/
create(thing: string, data: any): Promise<any> {
   [...]
}
```

And an example of how it render in VScode intelliSense :

Code_K9tb9PS8z2

michaelprichardson:
@Trystan-SA Great suggestion! Will look into adding it for the `Surreal` class too

oberbeck:
nice!

Just one side note - do you think its necessary to **re**state the types in the docs?

oberbeck:
Hi everyone, as you might have seen, we are having the same issues in the various JS repositories (https://github.com/surrealdb/surrealdb.deno/issues/5). What do you think of streamlining our efforts into one common base?

PR #12, opened 2022-09-17
Eliminate the need for `nanoid` module

jareer12:
I made it so you don't need to install an entire module for a single string.

Edit: I don't think you need to use `nanoid` since the IDs don't need to be cryptographically secure.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,6 @@
       "version": "0.3.1",
       "license": "MIT",
       "dependencies": {
-        "nanoid": "^4.0.0",
         "ws": "^8.8.1"
       },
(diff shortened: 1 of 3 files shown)


trs:
Alternately, we could use [`crypto.randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID), available in browers and node since v16.7.0.

gotjoshua:
there are some pretty lightweight hashing functions, that could also be considered to make the ids content specific

eg:  
https://github.com/bryc/code/blob/master/jshash/experimental/cyrb53.js

PR #14, opened 2022-09-17
Merge community requests

### LINKED ISSUE #1: javascript
no types. if only it was written in typescript in the first place

### LINKED ISSUE #11: Vite falling back to node ws (doesn't exist) on client side
# Reproducing

- Create a vite project
- Use surrealdb.js
- See the Buffer not defined error
- Add vite polyfill for buffer
- See the class is not defined

### LINKED ISSUE #13: Consider using isomorphic-ws, so there's no runtime check needed
[As suggested by](https://github.com/surrealdb/surrealdb.js/issues/1#issuecomment-1249313840) @Choooks22  isomorphic-ws could be a better option than ws with the homebrew environment checking.

Opening a separate issue to pull this out of the typescript thread as @tobiemh [indicated they wanted to look into it a bit more](https://github.com/surrealdb/surrealdb.js/issues/1#issuecomment-1227830485).

Choooks22:
My bad if the pr is too big orz.
Includes a lot of my local development from discussions from the discord server.

### TypeScript

- Includes mostly complete types with generic support. Not 100% coverage since the entirety of features has not been implemented yet.
  - Resolves #1
  - Closes #5
  - Resolves surrealdb/surrealdb.deno#3
  - Closes surrealdb/surrealdb.deno#4
- Minimal runtime changes, only adding `Array.isArray()` checks for type narrowing.

### Toolchain

- Switched from single file bundling using webpack to one-to-one file transpilation using rollup.
  - Resolves #11
- Build includes compatibility for Deno, Node, and Browser.
  - Resolves #12
  - Resolves surrealdb/surrealdb.deno#5
  - Closes surrealdb/surrealdb.deno#6

### Other

- Replaced `nanoid` with autoincrementing id.
  - Resolves #12
- Inject `isomorphic-ws` for Node/Browser during build time.
  - Resolves #13


rollup.config.ts
```diff
@@ -0,0 +1,44 @@
+import replace from '@rollup/plugin-replace'
+import { defineConfig } from 'rollup'
+import rename from 'rollup-plugin-rename'
+import typescript from 'rollup-plugin-ts'
```
mathe42, line 4: I would propose to use esbuild instead of rollup-plugin-ts is much faster

have a look at rollup-plugin-esbuild for that

rollup.config.ts
```diff
@@ -0,0 +1,44 @@
...
+      preventAssignment: true,
+      values: { '.ts': '.js' },
+    }),
+    replace({
+      preventAssignment: false,
+      values: { '// %isomorphic-ws%': 'import WebSocket from "isomorphic-ws"' },
+      delimiters: ['', ''],
+    }),
+    // fixes bundled tslib becoming cjs because it's inside a node_modules dir
+    rename({ map: name => name.replace(/node_modules/, 'external') }),
```
mathe42, line 19: That is strange! Will have a look into it.

rollup.config.ts
```diff
@@ -0,0 +1,44 @@
+import replace from '@rollup/plugin-replace'
+import { defineConfig } from 'rollup'
+import rename from 'rollup-plugin-rename'
+import typescript from 'rollup-plugin-ts'
+
+export default defineConfig({
+  input: './src/index.ts',
+  plugins: [
+    replace({
```
mathe42, line 9: I understand this reassignment after thinking about it. please add a comment and combine with second replace plugin instance.

rollup.config.ts
```diff
@@ -0,0 +1,44 @@
...
+    {
+      dir: 'dist/cjs',
+      format: 'cjs',
+      exports: 'named',
+      interop: 'auto',
+      preserveModules: true,
+      preserveModulesRoot: 'src',
+      plugins: [
+        // allows cjs users to import cjs files inside our esm package
+        rename({ map: name => name.replace(/\.js$/, '.cjs') }),
```
mathe42, line 40: Don't think that this is needed....?

mathe42 (changes requested):
I have some changes I would suggest.

Types are over all OK but not perfect. I would take the time and use the types I did in https://github.com/surrealdb/surrealdb.deno/pull/6 to make them better AFTER this gets merged (for example better types for on, off etc.).

I think we lost the browser-build on the way but that can be added quite simple.

rollup.config.ts
Choooks22: That's just the default behaviour, each directory inside `node_modules` will be treated as a package, and by default all packages are commonjs if not defined by `package.json`

rollup.config.ts
Choooks22: All `.js` files inside esm packages (which this one is, since `type: module` is set) are treated as esm, even if its contents are cjs

rollup.config.ts
Choooks22: I just needed something that would resolve the `tsconfig.json` file, I'll take a look at it in the morning :+1:

rollup.config.ts
Choooks22: I need the second instance since it has different delimiters, and I'm not exactly sure either if things might/could break in the future if I also changed the delimiters for the extension reassignment.

I'll go add the comments since on second look it does look a bit weird :+1:

mathe42: Oh didnt saw that. Than it is fine....

rollup.config.ts
mathe42: Yes but we don't have any commonjs files...

rollup.config.ts
mathe42: Never mind rollup-pluigin-ts is better than I thought leave it in! No change required (as type-files get worse when useing esbuild)!

rollup.config.ts
Choooks22: It's for the commonjs output for `require()` compatibility

Choooks22:
> I think we lost the browser-build on the way but that can be added quite simple.

I've intentionally omitted the single file bundle output since pretty much all bundlers can handle multiple files, but yes in the case of directly using it using `<script src="">` wouldn't work.

mathe42:
All of the above is resolved for me (only browser-build remains):

---
An browser build can be added if in rollup.config.ts the defineConfig recives an array with the current config and additionaly

```ts
{
    input: './src/index.ts',
    plugins: [
      replace({
        preventAssignment: true,
        values: { '.ts': '.js' },
      }),
      typescript(),
    ],
    output: [
      {
        dir: 'dist/browser',
        format: 'esm'
      },
    ],
  }
```
It might be a good idea to add a minifier (terser or something similar).

---


For better deno support it might be a good idea to add into the root of this reposetory a file mod.ts with

```ts
export { default } from "./src/mod.ts";
```

mathe42:
> > I think we lost the browser-build on the way but that can be added quite simple.
> 
> I've intentionally omitted the single file bundle output since pretty much all bundlers can handle multiple files, but yes in the case of directly using it using `<script src="">` wouldn't work.

Nope as we have import of `isomorphic-ws` will not work.  I want to use it with `<script src="..." type="module"></script>` as this will allow fast prototyping!

gotjoshua:
this looks amazing! looking forward to test it out!

package.json
```diff
@@ -19,27 +17,27 @@
   },
   "type": "module",
   "exports": {
-    "import": "./index.js",
-    "require": "./dist/lib/index.cjs"
+    ".": {
+      "import": "./dist/esm/index.js",
```
mathe42, line 21: Change to 
```
"import": {
  node: "./dist/esm/index.js",
  default: "./dist/browser/index.js"
}
```

package.json
```diff
@@ -2,12 +2,10 @@
   "name": "surrealdb.js",
   "version": "0.3.1",
   "description": "Javascript driver for SurrealDB",
```
mathe42, line 4: Add fields
```
  "unpkg": "dist/browser/index.js",
  "jsdelivr": "dist/browser/index.js",
```
so cdn willchoose the correct file to serve when file is not specified.

mathe42 (changes requested):
N/N

mathe42:
Sorry know I get how isomorphic-ws works. Sorry browser build might be not needed still looking into it...

mathe42:
> Sorry know I get how isomorphic-ws works. Sorry browser build might be not needed still looking into it...

The `import WebSocket from "isomorphic-ws"` does the following:

It loads 2 different file depending on use case node / browser.

For node it exports the `ws`-package.
For Browser it exports the `WebSocket` global. 

So it is more easy to create a bunde without the import (-> Browser-Build) and one with the import directly from `ws` (Node-ESM build + Node-CJS build).

With that change we would support:

1. Deno
2. Browser direct import from cdn for prototyping (as `import Surreal from 'https://unpkg.com/suirrealdb.js@x.x.x'`)
3. Browser with bundler (with the browser build)
4. Node as cjs
5. Node as esm

One thing I don't understand why don't we compile the code into a single file (so we have node.esm.js, node.cjs.js, browser.esm.js, types.d.ts). If the answer is better debugging then we can just enable source-maps...

Choooks22:
> One thing I don't understand why don't we compile the code into a single file

That was a deliberate change in my part, since there were discussions in the discord to be able to use either the ws or http client separately without using both with `Surreal`.

Having it separate like this means we can easily add a `surrealdb.js/http` and `surrealdb.js/ws` export.

Choooks22:
> For better deno support it might be a good idea to add into the root of this reposetory a file mod.ts with

Ah yeah I'm aware about deno using `/mod.ts` but I just completely missed that, good catch! :+1:

mathe42:
> > One thing I don't understand why don't we compile the code into a single file
> 
> That was a deliberate change in my part, since there were discussions in the discord to be able to use either the ws or http client separately without using both with `Surreal`.
> 
> Having it separate like this means we can easily add a `surrealdb.js/http` and `surrealdb.js/ws` export.

OK than I'm fine with it. Looks like you gave it some thinking.

mathe42 (approved):
> 2\. Browser direct import from cdn for prototyping (as `import Surreal from 'https://unpkg.com/suirrealdb.js@x.x.x'`)

With the current implemetaion we don't have in my above list the server-prototyping-build. I thought about it and I'm fine to leave it out (for now). After this gets merged I would create an issue for that and then we can discuss if this is something we want to add. But I don't want this to be blocking because of this **small** inconvinience!

So from my side this is good to get merged! 🚀

src/index.ts
```diff
@@ -1,61 +1,126 @@
...
+  value: any;
+}
+
+export interface RemovePatch extends BasePatch {
+	op: 'remove';
+}
+
+export interface ReplacePatch extends BasePatch {
+	op: 'replace';
+  value: any;
```
btmnk, line 26: I think some formatting tools could help to avoid inconsistencies in indentation and such.
Maybe adding eslint & prettier to the library makes sense?

src/index.ts
```diff
@@ -1,61 +1,126 @@
...
+let singleton: Surreal;
+
+interface BasePatch {
+	path: string;
+}
+
+export interface AddPatch extends BasePatch {
+  op: 'add';
+  value: any;
+}
```
btmnk, line 18:
```suggestion
export type PatchValue = string | number | null | undefined | Record<string, unknown> | Array<PatchValue>;

export interface AddPatch extends BasePatch {
  op: 'add';
  value: PatchValue;
}
```

Not sure if this is helpful or might introduce problems but it could help to avoid putting stuff in a modify payload that shouldn't be there, e.g. a function?

src/index.ts
```diff
@@ -401,61 +466,61 @@ export default class Surreal extends Emitter {
...
 
-	#result(res, resolve, reject) {
+	#result<T>(res: Result<T>, resolve: (value: T) => void, reject: (reason?: any) => void): void {
 		if (res.error) {
 			return reject( new Error(res.error.message) );
 		} else if (res.result) {
 			return resolve(res.result);
 		}
-		return resolve();
+		return resolve(undefined as T);
```
btmnk, line 475:
```ts
#result<T>(res: Result<T>, resolve: (value: T | undefined) => void, reject: (reason?: any) => void): void {
  if (res.error) {
	  return reject( new Error(res.error.message) );
  } else if (res.result) {
	  return resolve(res.result);
  }
  return resolve(undefined);
}
```
(Sorry, github didn't let me do a code suggestion, see the changes of the resolve type)

I'm not sure if `undefined as T` is even allowed in typescript?
Same change should be done for `#output`.

btmnk (commented):
Since I'm interested in surrealdb and wanted to get familiarized with this library a bit I had a look over these changes and added some comments for stuff I think could help.

(Sorry in advance if somethings are wrong or useless, I'm not that familiar with the library and surrealdb yet)

src/index.ts
mathe42: I would say useing `deno fmt` would be more fitting (but totaly agree!)

src/index.ts
mathe42: I think here is any the better fit. (I personaly have no problem with any in the right places!). I think `PatchValue` is not exposed API so it should be fine either way.

src/index.ts
mathe42: function parameter should be changed to `resolve: (value?: T) => void`.

src/classes/emitter.ts
```diff
@@ -1,15 +1,18 @@
+// deno-lint-ignore-file no-explicit-any
```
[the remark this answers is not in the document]
mathe42: I don't like this. `isNotifyEvent` is only used once so I don't see the need for this helper!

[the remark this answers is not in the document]
btmnk: Ah indeed, I think I chose a bad example there. My point was mostly to make sure the `e` in the callback is strongly typed to something instead of `any` to make working with it in the future more transparent. But I agree, it's probably not worth the overhead as it is only one unique case..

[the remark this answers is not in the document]
mathe42: I totaly agree in https://github.com/surrealdb/surrealdb.deno/pull/4 for the deno repo (wich is replaced by this PR) I added better types for the events. As said before after this gets merged I will have a look and check what to optimize etc. 

But I didn't want to add it here now as this will hold this PR back for merging. (Same for formater below).

src/index.ts
btmnk: Ah I see, I figured it would affect the way how `db.modify` would be used.
e.g. from the examples:
```
let person = await db.modify('person:tobie', [
	{ op: "replace", path: "/settings/active", value: false },
	{ op: "add", "path": "/tags", "value": ["developer", "engineer"] },
	{ op: "remove", "path": "/temp" },
]);
```
So it would assist in what to put into "value" even if it's just a type that contains basically every primitive it would give the developer a sense of "I can safely put anything there".
But I see that this might be unnecessary overhead when `any` works fine too since I think the only "wrong" type one could put there are functions, right?

mathe42: I would say to set it to any for know and I'll have a look into it when this is merged.

src/classes/live.ts
```diff
@@ -1,16 +1,17 @@
-import Emitter from "./emitter.js";
+import Surreal, { Result } from '../index.ts';
```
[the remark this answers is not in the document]
Choooks22: Whoops nope, I didn't setup linting yet so didn't notice

src/classes/emitter.ts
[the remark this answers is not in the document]
Choooks22: I disabled the checks since I didn't want to add any more runtime changes since that isn't the goal of the pr (which already covers one too many scopes).

I'll be up for improving types/runtime behaviour in later prs :+1:

mathe42:
@Choooks22 can you add `"types": "./dist/types/index.d.ts",` to readme currently types are not recogniced...

src/index.ts
Choooks22: I took the types from `@types/json-patch` as referenced in b614b18, didn't bother with making a perfectly typed system yet.

src/index.ts
Choooks22: `resolve: (value?: T) => void` had issues where ts didn't like that it could accept `value: undefined | T` instead of just `T`.

In the future this could be easily switched to using async/await which would completely remove the need for this type.

Choooks22:
> @Choooks22 can you add `"types": "./dist/types/index.d.ts",` to readme currently types are not recogniced...

Sure, but I wonder how far we should keep the `types` field, since typescript's new `moduleResolution: node16` can resolve the new types from `exports` field.

There's a hack using `typesVersions` to resolve subpath import types for compatibility, but I really don't like it since it feels really janky...

mathe42:
Ah 

> > @Choooks22 can you add `"types": "./dist/types/index.d.ts",` to readme currently types are not recogniced...
> 
> Sure, but I wonder how far we should keep the `types` field, since typescript's new `moduleResolution: node16` can resolve the new types from `exports` field.
> 
> There's a hack using `typesVersions` to resolve subpath import types for compatibility, but I really don't like it since it feels really janky...

Nice that works thanks! Then never mind - but a info in the readme would be a good idea!

Choooks22:
@mathe42 about the cdn build, are there any other special setup needed to upload to a cdn like unpkg / cdnjs / etc?

mathe42:
> @mathe42 about the cdn build, are there any other special setup needed to upload to a cdn like unpkg / cdnjs / etc?

No just the push to npm is enough. The only thing is that you might need to set a browser, unpkg or jsdelivr field in package.json see https://unpkg.com/vue@3.2.39/package.json for example. (I think for us only browser field would be needed)

src/utils/guid.ts
```diff
@@ -0,0 +1,5 @@
+let id = 0
+
+export default function(): string {
+	return (id = id + 1 % Number.MAX_SAFE_INTEGER).toString()
```
mathe42, line 4: `id + 1` should be `(id + 1)`

tobiemh:
Hi @Choooks22 @mathe42 @gotjoshua @btmnk, sorry for pinging you all, but wanted to get this merged as quickly as possible.

Couple of questions...

1. Obviously this will enable use in Deno too. Are we therefore removing the https://deno.land/x/surrealdb@v0.2.1 library, and would this repo be deployed to Deno? What changes would be required there?
2. Is this ready to be merged as it is at the moment?
3. Just confirming - does it work in Deno/Nodejs/Browser and with `require()` and `import ...` syntax?
4. To plan on a future release of this merge, what are the changes necessary for deployment to the different package systems NPM/Deno/CDNs...?

mathe42:
2\. My opinion yes (only my last comment in guid.ts should be fixed).
3\. Deno, Node, Browser with bundler works. Only import via CDNs are not properly supported.

To 1\. and 4\.:

1. Deno: if you can just use the webhookthat is setup for the deno repo and add it to this repo.
2. Node + Browser with bundler: Just publish as npm package (as before!).
3. Browser CDN currently not supported but with the publish to npm this would be done.

So only change is that you should change is the webhook for deno.

Choooks22:
> Obviously this will enable use in Deno too. Are we therefore removing the https://deno.land/x/surrealdb@v0.2.1 library, and would this repo be deployed to Deno? What changes would be required there?

The deno.land module can still be up, just pointing to this repo.

> Is this ready to be merged as it is at the moment?

Fixed the guid.ts error, and the changes mentioned in the discord are mostly type-related which we'd be focusing on next.

> Just confirming - does it work in Deno/Nodejs/Browser and with `require()` and `import ...` syntax?

Yes, separate esm and cjs builds are working.

Browsers needs bundlers to work, which is not unexpected since this is an npm package after all.

CDN build for use without bundlers should be automatically available after publish as mentioned by @mathe42.

> To plan on a future release of this merge, what are the changes necessary for deployment to the different package systems NPM/Deno/CDNs...?

I think the only other possible change would be to ship our own single-file bundle for CDNs, which this pr currently doesn't do.

btmnk:
> Hi @Choooks22 @mathe42 @gotjoshua @btmnk, sorry for pinging you all, but wanted to get this merged as quickly as possible.
> 
> Couple of questions...
> 
> 1. Obviously this will enable use in Deno too. Are we therefore removing the https://deno.land/x/surrealdb@v0.2.1 library, and would this repo be deployed to Deno? What changes would be required there?
> 2. Is this ready to be merged as it is at the moment?
> 3. Just confirming - does it work in Deno/Nodejs/Browser and with `require()` and `import ...` syntax?
> 4. To plan on a future release of this merge, what are the changes necessary for deployment to the different package systems NPM/Deno/CDNs...?

Thanks for the mention but I was only nitpicking in this PR; In general it looked good to me.
I tested it with Node + import and it worked fine.
(Important setting for tsconfig.json is to set moduleResolution to Node16 or NodeNext but nothing exceptional apart from that)

mathe42:
> CDN build for use without bundlers should be automatically available after publish as mentioned by @mathe42.

Small addition to that: Yes the CDN will pickup the npm module and we get the packages but CDN has no use for browsers because of the `isomorphic-ws` package that is not resolved. But as said to fix that we can have an extra single-file bundle for that (maybe also minified).

But as said I think that would be good for an extra PR so this can get merged!

mathe42 (approved):

tobiemh:
@Choooks22, @mathe42, to change the source repository on [deno.land](https://deno.land/x/surrealdb@v0.2.1) I need to contact Deno themselves.

I can't just switch the repository from `surrealdb.deno` to `surrealdb.js`.

Does anyone know a good person to contact on Deno Discord or elsewhere?

PR #21, opened 2022-09-18
feat: add linter + formater as github action

mathe42:
I use the same config (+ tests) for own project.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,12 +1,12 @@
 {
   "lint": {
     "files": {
-      "include": ["src/", "./mod.ts"]
+      "include": ["src/", "./mod.ts", "./rollup.config.ts"]
     }
   },
   "fmt": {
     "files": {
-      "include": ["src/", "./mod.ts"]
+      "include": ["src/", "./mod.ts", "./rollup.config.ts"]
     }
   }
 }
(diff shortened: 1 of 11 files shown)


oberbeck (approved):

VixieTSQ:
Wonder if it would be worth testing in bun, node, and deno because we support all of them and therefore should confirm it works for all of them.

mathe42:
@VixieTSQ I don't agree as this is (currently) only linting and formating....

mathe42:
@tobiemh ready for review (and merge?). I use the same config in my own project so it should be fine. It currently doesn't run as it is just in the fork repo.

deno.json
```diff
@@ -0,0 +1,12 @@
+{
+  "lint": {
+    "files": {
+      "include": ["src/", "./mod.ts", "./rollup.config.ts"]
+    }
+  },
+  "fmt": {
```
tobiemh, line 7: @mathe42 can we use tabs here? And then re-fmt the PR?

```js
"options": {
      "useTabs": true,
      "lineWidth": 80,
      "indentWidth": 4,
      "singleQuote": false,
 }
```

mathe42: Done!

VixieTSQ:
> @VixieTSQ I don't agree as this is (currently) only linting and formating....

Ah sorry. Thought this included tests.

PR #27, opened 2022-09-19
Typescript Fixes

minerj101:


src/classes/pinger.ts
```diff
@@ -1,5 +1,5 @@
 export default class Pinger {
-	#pinger?: ReturnType<typeof setTimeout>;
+	#pinger?: NodeJS.Timer | number;
```
oberbeck, line 2: we should be environment generic as we support not only NodeJS but also Deno and so on...

mathe42: Yes should be ReturnType<...>

mathe42:
Best setup:
1. Use VSCode
2. Install Deno-Extension
3. Install Deno

=> All errors are gone

mathe42:
And yes we need the `.ts` extensions in the files! This repo will be pushed with no change to `deno.land/x/surrealdb`

mathe42:
@minerj101 Thanks for the PR!

The only change that doesn't break anything is the tsconfig. But I am not a big fan of that much comments. So I think this will not get merged.

Could you close this PR?

mathe42:
@minerj101 @tobiemh can be closed

PR #29, opened 2022-09-19
fix: pinger type

mathe42:
This is a type where Node, Browser, Deno have different types so set it here to any.


Choooks22:
I used `ReturnType<typeof setTimeout>` since yes browser and node have different signatures (browser is just plain `number` while node is `NodeJS.Timeout`, but we don't have `NodeJS` ns from deno), but I have no issues to continue with this since it's internal.

mathe42:
For me with that type the build fails. (maybe because dom types and node types collide there)

@tobiemh ready to merge and ready to publish to npm

Choooks22 (approved):

PR #33, opened 2022-09-19
feat: improve emitter class

Choooks22:
- Slightly improves node.js `EventEmitter` interop for current methods.
- Fixes memory leak by removing empty arrays.
- Added `Emitter.once()` for converting once listeners to promises.
- Added support for event maps.

### DIFF
--- a/src/classes/emitter.ts
+++ b/src/classes/emitter.ts
@@ -83,7 +83,7 @@ export default class Emitter<Events extends EventMap = EventMap> {
 	): this;
 	once(eventName: EventName, listener: Listener): this {
 		this.on(eventName, function once(...args: any[]) {
-			this.removeListener(eventName, listener);
+			this.removeListener(eventName, once);
 			listener.apply(this, args);
 		});
 		return this;
--- a/src/classes/emitter.ts
+++ b/src/classes/emitter.ts
@@ -2,19 +2,33 @@
 export type EventName = string | symbol;
 export type Listener = (this: Emitter, ...args: any[]) => void;
 
-export default interface Emitter {
+export type EventMap = Record<EventName, any>;
+
+export default interface Emitter<Events extends EventMap = EventMap> {
+	addListener<T extends keyof Events>(
+		eventName: T,
+		listener: (this: this, ...args: Events[T]) => void,
+	): this;
 	addListener(eventName: EventName, listener: Listener): this;
+	off<T extends keyof Events>(
+		eventName: T,
+		listener: (this: this, ...args: Events[T]) => void,
+	): this;
 	off(eventName: EventName, listener: Listener): this;
 }
 
+export function once<T extends EventMap, K extends keyof T>(
+	emitter: Emitter<T>,
+	eventName: K,
+): Promise<T[K]>;
 export function once(emitter: Emitter, eventName: EventName): Promise<any[]> {
 	return new Promise((res) => {
 		emitter.once(eventName, (...args) => res(args));
 	});
 }
 
-export default class Emitter {
-	#events: Map<EventName, Listener[]> = new Map();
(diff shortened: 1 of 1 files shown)


mathe42:
Could we agree to use //@ts-expect-error instead of //@ts-ignore ?

Choooks22:
> Could we agree to use //@ts-expect-error instead of //@ts-ignore ?

I don't see why not

src/classes/emitter.ts
```diff
@@ -1,49 +1,110 @@
...
+	eventName: K,
+): Promise<EventsOf<T>[K]>;
+export function once(emitter: Emitter, eventName: EventName): Promise<any[]> {
+	return new Promise((res) => {
+		emitter.once(eventName, (...args) => res(args));
+	});
+}
+
+export default class Emitter<Events extends EventMap = EventMap> {
+	#events = new Map<EventName, Listener[]>();
```
mathe42, line 35: We can have a better type if using an object.

```ts
class X<Events> {
  #events = {
     [K in keyof Events]?: ((...args: Events[K]) => void)[] = {}
  }
}
```

src/classes/emitter.ts
```diff
@@ -1,49 +1,110 @@
 // deno-lint-ignore-file no-explicit-any
+export type EventName = string | symbol;
 export type Listener = (this: Emitter, ...args: any[]) => void;
```
mathe42, line 3: I would propose to change it to
```ts
export type Listener<T = any[]> = (...args: T) => void
```

Why do we have `this` here. The listener has not to be bound to the emitter

mathe42 (changes requested):
Some thoughts... 

If we change both we can get rid of the `// @ts-expect-error`...

src/classes/emitter.ts
mathe42: See also https://github.com/mathe42/surrealdb.deno/blob/82dee508405f5662beecaa4624b0dff9bc866afe/src/classes/emitter.ts how I did it.

src/classes/emitter.ts
mathe42: See also https://github.com/mathe42/surrealdb.deno/blob/82dee508405f5662beecaa4624b0dff9bc866afe/src/classes/emitter.ts

src/classes/emitter.ts
```diff
@@ -1,49 +1,110 @@
...
-	emit(e: string, ...args: any[]) {
-		if (typeof this.#events[e] === "object") {
-			this.#events[e].forEach((func) => {
-				func.apply(this, args);
-			});
+	emit<T extends keyof Events>(eventName: T, ...args: Events[T]): this;
+	emit(eventName: EventName, ...args: any[]): this {
+		const listeners = this.#events.get(eventName);
+		if (listeners !== undefined) {
+			listeners.forEach((listener) => listener.apply(this, args));
```
mathe42, line 96: why use apply. This will break a lot of code I think `listener(...args)` would be better.
But if we leave it yes we would need the this evrywhere....

src/classes/emitter.ts
Choooks22: Because it was part of the original code and it's how node's `EventEmitter` handles it

```ts
import { EventEmitter } from 'node:events'

const ee = new EventEmitter()

ee.on('foo', function() {
  console.log(this) // EventEmitter {}
})

ee.emit('foo')
```

mathe42:
I created a PR with some type optimizations https://github.com/Choooks22/surrealdb.js/pull/1

mathe42:
And added types for all subclasses.

tobiemh:
Just checking @mathe42 and @Choooks22 that this PR is to be closed now?

mathe42:
Yes can be closed

PR #35, opened 2022-09-19
deno: add default to export in /mod.ts

mathe42:
This fixes the currently broken /mod.ts...


Choooks22 (approved):
Can't believe I missed this one :sweat:

PR #44, opened 2022-09-23
Add emitter types

oberbeck:


mathe42:
Dublicate of #33

oberbeck:
> Dublicate of #33

ouch... I am amazed how similar the solutions are though 🤓

May the better one win ✌️

src/classes/emitter.ts
```diff
@@ -1,41 +1,72 @@
...
+> {
+	#events: {
+		[TKey in keyof TEmitterEvent]?: Listener<this, TEmitterEvent[TKey]>[];
+	} = {};
+
+	on<TKey extends keyof TEmitterEvent>(
+		e: TKey,
+		func: Listener<this, TEmitterEvent[TKey]>,
+	) {
+		if (!Array.isArray(this.#events[e])) {
```
mathe42, line 28: Why this check either it is `undefined` or an Array. (empty arrays doesn't matter)

oberbeck: Yup, that's what I thought, too. But somehow with this Deno/Node setup it would still complain... hence I added the `!`

Issue #32, opened 2022-09-19
strict typing internals


oberbeck:
```
Each request is an object (SurrealQL object) which needs
1.  id field (random id, no necessary length, just a string)
2. method field (a string)
3. params field (an array of values)...

method can be... ping info use signup signin authenticate invalidate authenticate kill live let set query select create update change modify delete

params is an array with different requirements depending on the method...

ping empty array (ping is just used to keep connection open over some proxies - maybe set a poller to call this every 30 seconds or so)...
info empty array (used to get the current $auth object of the currently authenticated user - I'll explain better in due course)
use array with two string values [NS, DB],
    NS if set (and not undefined) will set the NAMESPACE
    DB if set (and not undefined) will set the DATABASE
signup takes one value which is an object (can accept any keys in the object)
signin takes one value which is an object (can accept any keys in the object)
authenticate takes one value which is a string (a JWT token)
invalidate empty array
kill takes one value which is a string (specifically a uuid)
live takes one value which is a string (the name of a table)
let takes two values, the first a string (the param name), the second any value (treated like JSON)
set alias for let
query takes two values, the first a string (SurrealQL query text), the second an object with any keys (parameters to be used in the sql query)
        returns an array of query results, each with an array of records (because you can send multiple statements in a single query)
select takes one value (the name of a table or a record id)
        returns an array of records (it takes just the first query resultset)
create takes two values, the first is the name of a table or a record id, the second is either undefined or an object (the record content)
        returns an array of records (it takes just the first query resultset)
update takes two values, the first is the name of a table or a record id, the second is either undefined or an object (the record content)
        returns an array of records (it takes just the first query resultset)
change takes two values, the first is the name of a table or a record id, the second is either undefined or an object (the record content)
        returns an array of records (it takes just the first query resultset)
modify takes two values, the first is the name of a table or a record id, the second is an array of JSONPatch objects to apply to the records
        returns an array of record DIFFS after thenchanges were made (it takes just the first query resultset)
delete takes one value,  the name of a table or a record id
        returns an array of records (which will be empty)
```

source: https://discord.com/channels/902568124350599239/1013527402674139268/1021058706080813166

oberbeck:
FYI, I started working on it

oberbeck:
I have created two example implementations and was wondering which one you would prefer:

https://github.com/oberbeck/surrealdb.js/commit/f76aa58d4f016376b1684dd9dae33b4fbbeef32a
[...]

PR #47, opened 2022-09-24
Feat/add request typings

oberbeck:
* https://github.com/surrealdb/surrealdb.js/issues/32

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -1,0 +1,151 @@
+export enum SurrealQLMethod {
+	ping = "ping",
+	info = "info",
+	use = "use",
+	signup = "signup",
+	signin = "signin",
+	authenticate = "authenticate",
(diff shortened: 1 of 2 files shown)


mathe42:
I prefer request.ts with `enum SurrealQLMethod` removed. Instead use the strings directly.

oberbeck:
> I prefer request.ts with `enum SurrealQLMethod` removed. Instead use the strings directly.

What's the matter with the enum? You can still use strings directly if you want.

```typescript
const example0:SurrealQLMethod = SurrealQLMethod.ping; // valid
const example1:SurrealQLMethod = "ping";               // also valid
```

Personally, I prefer to support both, as enums are more reliable when it comes to refactoring.

kearfy:
Closing as stale

PR #50, opened 2022-09-25
Feat/add meta info

oberbeck:


mathe42:
I like the following patterns more:

```ts
function anyFn(arg1, arg2, db: Surreal = Surreal.instance) {
  // do stuff
}
```
Or
```ts
function anyFn(db: Surreal = Surreal.instance) {
  return (arg1, arg2) => {/*do stuff*/}
}
```

PR #53, opened 2022-09-27
#ws type

matigumma:
for NodeJS
correct type to infer on methods from Emitter
like this.#ws.on(...

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -119,7 +119,7 @@ export default class Surreal extends Emitter {
 	// Properties
 	// ------------------------------
 
-	#ws!: Socket;
+	#ws!: Socket & Emitter;
 
 	#url?: string;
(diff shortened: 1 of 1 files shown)


mathe42:
Why is this on a private field needed?

oberbeck:
Are you sure there isn't something strange with your local setup/environment?
`Socket extends Emitter` and therefore `#ws` has all of `Emitter`'s public fields. This should also hold true for any TypeScript setup AFAIK. Not sure what you mean with NodeJS here.

Further, this project can't be run with NodeJS directly anyway.

https://github.com/surrealdb/surrealdb.js/blob/ce949aeddd2b451b3b7b473705e62fbbc58e095b/src/classes/socket.ts#L23

image

PR #55, opened 2022-09-28
Idea: QueryBuilder

oberbeck:
I have started to tinker around with a basic query builder example.

The objective I am thinking about is a tooling which gets a DB model as its generic input (ie. the output of https://github.com/surrealdb/surrealdb/issues/248) and then lets you generate SELECT SurrealQL queries with a typed result.

### DIFF
--- a/src/classes/query-builder.ts
+++ b/src/classes/query-builder.ts
@@ -1,0 +1,245 @@
+//#region "DotPath" Type Helper
+
+//#region @source: https://stackoverflow.com/a/58436959/4295410
+
+type Prev = [
+  never,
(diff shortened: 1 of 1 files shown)


MatthewAry:
Regarding this. I have always liked how https://knexjs.org/ builds it's queries. If a query builder feature were to be implemented, I suggest:
- It be inspired by knex
- That it requires explicit imports to use because some people might want to just manually build their queries.

ochrstn:
Take a look at this type-safe typescript SQL query builder also: https://github.com/koskimas/kysely

PR #56, opened 2022-09-30
feat: add transaction

mathe42:
As disccused in https://discord.com/channels/902568124350599239/1018618253695795261/1025311316220588102

I think it is a good syntax for it.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -605,6 +605,39 @@ export default class Surreal extends Emitter<
 		return;
 	}
 
+	/**
+	 * Run queries inside a transaction!
+	 *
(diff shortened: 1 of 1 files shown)


kearfy:
@mathe42 Thanks for your PR, I will have to close it though. We intend the keep the client drivers as bare bones as possible and to only implement methods that are also implemented in the websocket protocol.

Once again, thanks for your contributions! We really appreciate it :)

tobiemh:
This change wouldn't actually work anyway at the moment, until we support client-based transactions. Currently only atomic transactions are supported.

PR #57, opened 2022-10-04
feat: rework sending, connection status

mathe42:
This PR changes the following:
1. make index.ts more readable by creating the id etc. in the `#send` helper
2. If `.connect` is not called before any function we throw with an error message (https://github.com/mathe42/surrealdb.js/blob/168da41e2a9fe8b07da6cc49a7529a55608e612f/src/index.ts#L294)
3. Expose state of websocket. For possible states see https://github.com/mathe42/surrealdb.js/blob/168da41e2a9fe8b07da6cc49a7529a55608e612f/src/classes/socket.ts#L20
4. Combine some eventListeners into a single one in socket.ts
5. Some small refactoring of the output handler.

### DIFF
--- a/src/classes/socket.ts
+++ b/src/classes/socket.ts
@@ -33,7 +33,7 @@ export enum ConnectionState {
 	/**
 	 * Connection open
 	 */
-	OPEN
+	OPEN,
 }
 
 export default class Socket extends Emitter<{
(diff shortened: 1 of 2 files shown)


mathe42:
@tobiemh ready for review

kearfy:
Hey @mathe42, thanks for your contribution! I will unfortunately be closing this as I rebuilt the complete websocket and emitter in #95.

PR #63, opened 2022-11-29
refactor(types): fix query generic

sasial-dev:
This PR fixes the type so that the result is correctly returned when using a type generic.


sasial-dev:
I just did this on github.com quickly, so let me know if there's any failing tests with this change! :)

mathe42:
lgtm - but is a breaking change so bump to 0.6 is required at release.

kearfy (approved):
Thanks! Had this thought a while ago too :)

Issue #68, opened 2023-01-24
Surreal.wait() should return an error if ws connection is reset by the client.


sachinbhutani:
Multiple components and pages are using single client websocket connection on my app. if a browser event (like refresh or back) resets the client connection, we get a hard typeerror from surreal.wait() function 

TypeError: Cannot read properties of undefined (reading 'ready')
at Surreal.wait (index.ts:257:14)

    let DB = new Surreal()
    await DB.connect("http://localhost:8000/rpc")

    await DB.signin ({user: "root", pass: "xxxx"})

    await DB.use("test","test")

    let info = await DB.query("INFO FOR DB;")

    console.log (info)

    DB = new Surreal() //To simulate browser refresh - other client events which may reset the connection from the client.

    info = await DB.query("INFO FOR DB;")

    console.log(info) // this statement dumps with `typeError

jareer12:
Don't pass `await` functions to `console.log()` method directly.
```ts
const info = await DB.query("INFO FOR DB;")
console.log(info)
```

You need parentheses to the `Surreal` constructor.
```ts
const DB = new Surreal()
```

PR #75, opened 2023-03-29
Use Surreal.wait() method to ensure connection to database

YutaUra:
I'm not sure if this modification is necessarily correct, but I hope it can be useful for something.

## Related

- https://github.com/surrealdb/surrealdb.js/issues/68


kearfy (approved):
Thanks for this @YutaUra! Looks good to me!

PR #77, opened 2023-04-02
Client can only signup to scope

kearfy:
The typing indicates that a client can signup as an instance, namespace, database and scope user. This is not true, as a client can only signup to a scope


kearfy:
Closing in favor or #78

PR #82, opened 2023-04-11
Bump to 0.6.0

kearfy:


kearfy:
Whoop, bump #major was definitely not a good name 🙈

PR #85, opened 2023-04-12
fix: match auth options

dvanmali:
Update the auth types to match the SurrealDB [verify types](https://github.com/surrealdb/surrealdb/blob/main/src/iam/verify.rs) while authenticating.


kearfy:
Heya! 

If I'm not mistaken, the TK and ID properties are fields that belong in a custom JWT token. They should not be passed to a signup/signin function.

Could I ask what you intention is with this PR?

-- Micha

kearfy:
Closing as stale

PR #104, opened 2023-05-08
Fix typings for `.select()`, `.create()`, `.update()`, `.change()`, `.modify()`, `.delete()`

### LINKED ISSUE #100: v0.7.0 returns `Promise<any>` for `select`, `create`, `update`, `change` and `modify`
All of the methods in the title now return `Promise<any>` even with generics specified

kearfy:
## What changes
After this PR you will need to explicitly specify a record ID:
```typescript
// Referencing an entire table stays the same:
await surreal.select('user');

// Referencing a specific record changes:
await surreal.select('user:tobie');       // Old
await surreal.select(['user', 'tobie']);  // New
```

## Why does it change
By requiring users to explicitely specific record IDs we:
- Fix #100 
- Get rid of hacky way of returning single object or array
- Sync implementation with rust API

## Breaking (sort of)
The way you reference a specific record ID in the `.select()`, `.create()`, `.update()`, `.change()`, `.modify()`, `.delete()` methods is breaking. The nice thing about the way this change is done is that if you don't change your way of doing things, you can only break the typing, which was already broken anyways. The actual code will still work :)


AlexFrid (approved):
LGTM

Issue #101, opened 2023-05-07
Surreal() constructor not working without url


sachinbhutani:
In earlier versions the Surreal constructor did not need any arguments. This allowed to declare a const like 
export const DB = new Surreal(); 

This is not working anymore.

kearfy:
Fixed by #105

Issue #94, opened 2023-05-04
Surreal.create returns an array when the id is randomly generated


moresdavidewayan:
```ts
import Surreal from "https://deno.land/x/surrealdb/mod.ts";

const db = new Surreal(URL);
const result = await db.create("person", {
	title: "Founder & CEO",
	name: {
		first: "Tobie",
		last: "Morgan Hitchcock",
	},
	marketing: true,
	identifier: Math.random().toString(36).substr(2, 10),
});
```
`result` should contain an object representing the newly created record.
An array of objects containing a single object is returned instead.

moresdavidewayan:
The easiest way to solve this is by adding an optional parameter to the `Surreal.#outputHanlder` function.
```ts
	#outputHandler<T>(
		res: Result<T>,
		thing: string,
		error: typeof Error,
		errormessage: string,
		single: boolean = false
	) {
		if (res.error) {
			throw new Error(res.error.message);
		}

		const isSingleThing = single || (thing && thing.includes(":"));
		if (Array.isArray(res.result) && res.result.length) {
			return isSingleThing ? res.result[0] : res.result;
		} else if ("id" in (res.result ?? {})) {
			return res.result;
		}

		throw new error(errormessage);
	}
```
Then on the `Surreal.create` function I'll add the additional parameter
```ts
	async create<T extends Record<string, unknown>>(
		thing: string,
		data?: T,
	): Promise<T & { id: Thing }> {
		const id = guid();

		await this.wait();
		this.#send(id, "create", [thing, data]);
		const [res] = await this.nextEvent(id);
		if (res.error) {
			throw new Error(res.error.message);
		}

		return this.#outputHandler(
			res,
			thing,
			PermissionError as typeof Error,
			`Unable to create record: ${thing}`,
			true
		);
	}
```


I don't like this approach since it makes code hard to maintain.

moresdavidewayan:
A better approach might be to create an enumerator type to represent the operation we are trying to perform.
```ts
enum DBOperation {
	SELECT,
	CREATE,
	UPDATE,
	CHANGE,
	MODIFY,
	DELETE
}
```

Then change  `Surreal.#outputHandler` so it can accept the operation as a parameter.
```ts
	#outputHandler<T>(
		operation: DBOperation,
		res: Result<T>,
		thing: string,
		error: typeof Error,
		errormessage: string,
	) {
		if (res.error) {
			throw new Error(res.error.message);
		}

[...]

PR #105, opened 2023-05-08
Fix for #101: Surreal() constructor not working without url

moresdavidewayan:
# Problem
Since #95 users couldn't instantiate a Surreal object without connecting to a running db instance (#101).
# Solution
`Surreal` constructor used to be implemented like this:
```ts
	constructor(url?: string, token?: string) {
		super();
		this.#url = url;
		this.#token = token;
		if (url) {
			this.connect(url);
		}
	}
```

I've changed the HTTPStrategy and WebSocketStrategy constructors to resemble the old constructor.
I've also added a new test to check if the create method for a record with a randomly generated id works correctly (#94)


kearfy:
Hiya, Thanks for this PR @moresdavidewayan. I was just about to open one myself! 😅 
Two points:
- Could you also make the url parameter optional in the src/types.ts file?
- I do not believe a test with a random ID to be relevant, as this is a functionality for surrealdb itself, not the library. 
  - The `.create()` function now always returns a single item by force. 
  - Typing for these functions is further improved by #104

moresdavidewayan:
Done

kearfy (approved):

PR #112, opened 2023-05-12
Live Queries implementation

kearfy:
# Adds back LQ functionality

Start a live query
```typescript
const uuid = await surreal.live<{
    id: string;
    prop: number;
}>("query...", (data) => {
    if (data.action === 'CLOSE") return;
    console.log(data.result);
});
```

Register another listener for an LQ
```typescript
await surreal.listenLive<{
    id: string;
    prop: number;
}>("uuid...", (data) => {
    if (data.action === 'CLOSE") return;
    console.log("Second listener! ", data.result);
});
```

Kill a LQ
```typescript
await surreal.kill("uuid...");
```

### Data in callback
```typescript
type close = {
    action: "CLOSE",
    detail: "SOCKET_CLOSED" | "QUERY_KILLED"
};

type create = {
    action: "CREATE";
    result: Record<string, unknown>;  // Generic type that was passed, always extends default type
};

type update = {
    action: "UPDATE";
    result: Record<string, unknown>;  // Generic type that was passed, always extends default type
};

type delete = {
    action: "DELETE";
    result: Record<string, unknown>;  // Generic type that was passed, always extends default type
};
```

### DIFF
--- a/src/library/SurrealSocket.ts
+++ b/src/library/SurrealSocket.ts
@@ -137,23 +137,25 @@ export class SurrealSocket {
 
 	async kill(query: string) {
 		if (query in this.liveQueue) {
-			this.liveQueue[query].forEach(cb => cb({
-				action: "CLOSE",
-				detail: "Query killed"
-			}));
+			this.liveQueue[query].forEach((cb) =>
+				cb({
+					action: "CLOSE",
+					detail: "Query killed",
+				})
+			);
 
 			delete this.liveQueue[query];
 		}
 
 		await new Promise<void>((r) => {
-			this.send('kill', [query], (_) => {
+			this.send("kill", [query], (_) => {
 				if (query in this.unprocessedLiveResponses) {
 					delete this.unprocessedLiveResponses[query];
 				}
 
 				r();
 			});
-		})
+		});
 	}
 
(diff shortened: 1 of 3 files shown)


kearfy:
So a complete message looks like this then @phughk?

```json
{
    "id": null,
    "method": "notify",
    "params": [
        {
            "uuid": "...",
            "action": "...",
            "result": { ... }
        }
    ]
}
```

I thought the uuid prop was query, I'll change that up!

phughk (approved):
lgtm, thanks!

phughk:
@kearfy the uuid prop is used to differentiate between live queries when they are received. The engine doesn't track connections, so it bubbles up all relevant LQ events. It's up to the consumer to differentiate. Likewise, LQ does not accept a uuid as that is generated server side - I think your API reflects that, but not entirely certain. It looks like it's accepted as an arg in the closure. Mentioning in case

kearfy:
Yep, I know! I thought that the `uuid` prop was called `query` as per previous examples that I've seen 😅 
But yeah, that array of params are all separate messages for potential different LQs if I understand it correctly. I do split those out 👍

kearfy:
As we already agreed on the messaging and the code has been tested on the current format of the messaging, we're taking the bet to merge a bit early to unblock others

PR #113, opened 2023-05-12
fix - fixed import of isomorphic-ws to work with nextjs server components

bibaswan-bhawal:
Fixed the import of `isomporhic-ws` for nodejs so that the library can be used within nextjs server components and apis


kearfy:
This partially solves the problem, but the build will fail for the browser package. Opening a PR in a second with a fix!

PR #134, opened 2023-06-22
Return the response from the info() method

zinyando:
## What is the motivation?

Return the current authenticated user form the `info()` method.

## What does this change do?

The `info()` method does not currently return anything, this MR returns the result from the method.


kearfy:
Heya, you'll need to update the types file and http strategy too. The info method would accept a generic type (as we don't know what the user record looks like)

kearfy:
Superseded by #135

Issue #133, opened 2023-06-21
TS improvements for Constructor and auth methods


itsezc:
Currently, you could do the following:

```ts
const db = new Surreal(PUBLIC_SURREAL_HOST, {
  db: PUBLIC_SURREAL_DB,
  ns: PUBLIC_SURREAL_NS
});
```

this is fully typed (and greatly appreciated), however when you then use methods such as `.signin` and `.signup` you're asked to pass in DB and NS (again):

```ts
await db.signin({
  user: username,
  NS: PUBLIC_SURREAL_NS,
  DB: PUBLIC_SURREAL_DB,
  SC: PUBLIC_SURREAL_SC
});
```

What I'd prefer to see:

```ts
await db.signin({
  user: username,
  SC: PUBLIC_SURREAL_SC
});
```

This is as the docs suggest on the README, currently I'm on version `0.8.0`

PR #141, opened 2023-07-15
Use existing values for defaults in signup method

PGilbertSchmitt:

## What is the motivation?

[Issue 133](https://github.com/surrealdb/surrealdb.js/issues/133) asked for a better experience withe the `signin` method, but I thought it would make sense to start with `signup` to see if my proposal is acceptable first. The motivation is that it is redundant to provide the namespace and database values to the `signup` method if they were already provided via the `use` method.

## What does this change do?

This PR updates the `Connection` interface's `signup` method so that the `NS` and `DB` parameters are optional. Using the `use` method beforehand can set these values on the connection strategy object, so we can populate the parameters accordingly.

I've tested my changes manually to ensure that for both the WS and HTTP strategies, the `NS`/`DB` parameters can be omitted from the `signup` method as long as they were previously passed to the connection via the `use` method, and that passing the values via the `signup` method still be used to override the `use` values.

This is related to, but does not directly solve, [issue 133](https://github.com/surrealdb/surrealdb.js/issues/133).


src/strategies/http.ts
```diff
@@ -103,10 +103,14 @@ export class HTTPStrategy<TFetcher = typeof fetch> implements Connection {
 	 * @param vars - Variables used in a signup query.
 	 * @return The authentication token.
 	 */
-	async signup(vars: ScopeAuth) {
+	async signup({
+		NS = this.http?.namespace,
+		DB = this.http?.database,
+		...rest
+	}: Partial<ScopeAuth> & Pick<ScopeAuth, "SC">) {
 		const res = await this.request<HTTPAuthenticationResponse>("/signup", {
```
kearfy, line 111: Could you do the isNil check here aswell?

src/types.ts
```diff
@@ -11,7 +11,7 @@ export interface Connection {
 	info?: <T extends Record<string, unknown> = Record<string, unknown>>() =>
 		Promise<T | undefined>;
 
-	signup: (vars: ScopeAuth) => Promise<Token>;
+	signup: (vars: Partial<ScopeAuth> & Pick<ScopeAuth, "SC">) => Promise<Token>;
```
kearfy, line 14: Instead of doing this picking and partially-ing, let's perhaps make the NS and DB props optional in the Auth types themselves? Should make things a bit more readable

kearfy (changes requested):
Heya, thanks for this! I added some comments, feel free to also implement this for the signin method :)

src/library/SurrealHTTP.ts
```diff
@@ -4,10 +4,11 @@ import { processUrl } from "./processUrl.ts";
 export class SurrealHTTP<TFetcher = typeof fetch> {
 	private url: string;
 	private authorization?: string;
-	private namespace?: string;
-	private database?: string;
 	private fetch: TFetcher;
 
+	_namespace?: string;
+	_database?: string;
```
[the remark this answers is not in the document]
PGilbertSchmitt: Holdover from before using getters, no problem

src/types.ts
PGilbertSchmitt: That's way better. I didn't want to start messing with your types right away.

src/strategies/http.ts
PGilbertSchmitt: :facepalm:

kearfy:
I further continued the implementation of this concept in #145 and will close this PR. Nevertheless, a big thank you for contributing! It is much appreciated

PR #143, opened 2023-07-17
Split signin method

PGilbertSchmitt:

## What is the motivation?

The `signin` method can't know in advance which authentication details to send to the server. The user could be a database or scoped user, in which case the `DB` and `NS` are necessary, but if the user is a namespace user, sending the `DB` will cause the authentication to fail. In order to make this process simpler for the user (type info, useful error messages).

## What does this change do?

This change splits the `signin` method into 4 different methods:
- `signin` remains for scoped users
- `namespaceSignin`
- `databaseSignin`
- `superUserSignin`

(This change also incorporates the changes from an [unmerged PR](https://github.com/surrealdb/surrealdb.js/pull/141))

- Manual testing
- TODO this this PR: Add unit tests.

- [Issue 133](https://github.com/surrealdb/surrealdb.js/issues/133)


kearfy:
Hiya! Eventhough this would allow to make the `NS` (and `DB`) properties optional for namespace and database authentication, it does not align with the rest of the client libraries and the websocket protocol. In #145, I made the `NS` and `DB` properties optional for specifically scope authentication which is most used besides root authentication.

Nevertheless, a big thank you for contributing! It is much appreciated

PR #168, opened 2023-09-14
Add missing JSON patch operations

meppu:
This PR aims to implement types for missing JSON patch operations, which can be found in the [surrealdb source code](https://github.com/surrealdb/surrealdb/blob/main/lib/src/sql/operation.rs).


kearfy (approved):
LGTM! Thanks!

Issue #203, opened 2023-12-14
Bug: db is not reconnecting


finalclass:
### Describe the bug

When surrealdb dies the library does not try to reconnect and it's not throwing (catchable) error.

### Steps to reproduce

1. Start surreal
2. Create new Surreal(), connect, signin, use...
3. stop surreal
4. start it again
5. try running some query using the `new Surreal()` instance.

This is the error:

```
error: Uncaught (in worker "") (in promise) InvalidStateError: readyState not OPEN
		this.ws?.send(JSON.stringify({ id, method, params }));
		         ^
    at WebSocket.send (ext:deno_websocket/01_websocket.js:327:13)
    at SurrealSocket.send (https://deno.land/x/surrealdb@v0.11.0/src/library/SurrealSocket.ts:133:12)
    at eventLoopTick (ext:core/01_core.js:197:13)
```

### Expected behaviour

The library either:
1. throws an error which can be caught
2. executuse "onError" hook
3. tries to reconnect

### SurrealDB version

1.0.0+20230913.54aedcd for linux on x86_64

### SurrealDB.js version

0.11.0

### Contact Details

s@finalclass.net

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

oliver-oloughlin:
Reproduced using Deno 1.40.0 + Surrealdb v1.2.0 + surrealdb.js v0.11.0. The part about uncatchable errors is a major issue for me, and makes the driver unusable in my use case. It seems WebSockets in Deno are unstable and tend to lose connection at random. This bug prevents the connection from being reestablished, and crashes my program as I am unable to catch the errors and reconnect manually.

akkie:
Same problem occurs with the latest beta version. Also, if I subscribe for the `ConnectionStatus.Error` or `ConnectionStatus.Disconnected` event, non of them gets triggered if SurrealDB is shut down. The node app gets killed because of the exception:

```
node:internal/deps/undici/undici:13178
      Error.captureStackTrace(err);
            ^

TypeError: fetch failed
    at node:internal/deps/undici/undici:13178:13
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
[...]

PR #220, opened 2024-03-30
fix: reject send promise upon catching inner error

oliver-oloughlin:
## What is the motivation?
Currently, errors that occur within `SurrealSocket.send()` are not surfaced properly, making them uncatchable. This leads to issues such as crashing a user's program if an error occurs, without letting them re-establish the database connection or continue execution gracefully.

## What does this change do?
This PR catches any inner errors, upon which the outer `send()` promise is rejected with the subsequent error. This allows for the error to be surfaced and caught.

I tested this change by running an infinite loop that would try to insert data into a running in-memory SurrealDB instance. The insert statement was surrounded by a try/catch block which would break the loop upon catching any errors and print a console.log message. While the loop was running, I would stop the SurrealDB instance and observe the results in the terminal. Before the change, the program would simply crash with the uncaught error. After the change, the loop would break gracefully and the console.log would be displayed, indicating the error had been caught and handled as expected. I also tried having the database re-establish the connection after a 5 second timeout instead of breaking the loop, where I would restart the database instance before the timeout would end. When doing this, the loop started running again, successfully inserting data.

These changes are related to #203. Ths PR does nothing directly to re-establish the database connection upon losing it, but it does surface otherwise uncatchable errors, allowing for the user to manually re-establish the connection.

- [✓] I have read the [Contributing Guidelines](https://github.com/surrealdb/surrealdb.js/blob/main/CONTRIBUTING.md)

### DIFF
--- a/src/strategies/websocket.ts
+++ b/src/strategies/websocket.ts
@@ -449,9 +449,11 @@ export class WebSocketStrategy implements Connection {
 		method: string,
 		params?: unknown[],
 	) {
-		return new Promise<U>((resolve) => {
+		return new Promise<U>((resolve, reject) => {
 			if (!this.socket) throw new NoActiveSocket();
-			this.socket.send(method, params ?? [], (r) => resolve(r as U));
+			this.socket
+				.send(method, params ?? [], (r) => resolve(r as U))
+				.catch((e) => reject(e));
 		});
 	}
 


oliver-oloughlin:
I might add that I think a better solution would be to restructure the current code somehow such that there isn't an unawaited promise when calling the `SurrealSocket.send()` method from within `WebSocketStrategy.send()`. However, this would certainly require more work and be a much more intrusive change. I believe the current proposed change does effectively address the issue with minimal changes to the code.

oliver-oloughlin:
I went ahead and created an alternative change branch which takes a different approach and doesn't leave the inner promise unawaited, here:
https://github.com/oliver-oloughlin/surrealdb.js/tree/bugfix-203-ensure-ws-errors-are-surfaced-2

Maybe worth considering?

mudlee:
I don't know how to upvote, but I'm upvoting implicitly here.

oliver-oloughlin:
@kearfy Hi, would you be able to give a review of this PR please? :)

itsezc (approved):

kearfy:
Hey everyone! Apologies for the delay here. We've been working on a refresh of the driver in a jsv2 branch, however I'll merge this first and get it out in a patch release. Thanks for this @oliver-oloughlin, great find!

oliver-oloughlin:
Seems the workflows are failing, but not sure if its related to this PR or not?

The linting error is about the `assert` keyword being deprecated, which should be replaced with `with`.
The tests fail when I run them locally on the main branch as well, without any changes applied.
There is also a type error in compile.ts when using `deno check` regarding the compiler options, where `lib: ["dom"]` should be `lib: ["DOM"]`.

Seems like there are some issues that maybe need fixing before this can be merged?
Or alternatively, if you are creating a refreshed driver anyway, just utilize what we have learned here when re-implementing it?

kearfy (approved):
LGTM!

PR #236, opened 2024-04-20
Fix live query UUID

tai-kun:
Thank you for the great work on this project.

## What is the motivation?

I was unable to subscribe to record changes using the live method. After researching, I found the following:

- Live query ID is verified by `z.string().uuid()`.
- In actuality, it is decoded as the UUID class, not a string.

## What does this change do?

- Make live queries work.
- Changing the string type to UUID prevents users from using it incorrectly.

Add three tests to `tests/integration/tests/live.ts`:

1. The test suite "live" checks that the `live` method returns a UUID class on the WebSocket engine and verifies that live queries work. Also, checks that it fails on the HTTP engine.
2. The test suite "unsubscribe live" checks whether a subscribed live query can be unsubscribed using the UUID class returned by the `live` method. As far as the HTTP engine is concerned, it doesn't support this, so we don't test anything.
3. The test suite "kill" checks whether a subscribed live query can be canceled using the UUID class returned by the `live` method. As far as the HTTP engine is concerned, it doesn't support this, so we don't test anything.

N/A

### DIFF
--- a/src/surreal.ts
+++ b/src/surreal.ts
@@ -14,6 +14,7 @@ import { Engine, HttpEngine, WebsocketEngine } from "./library/engine.ts";
 import { RecordId } from "./library/cbor/recordid.ts";
 import { Emitter } from "./library/emitter.ts";
 import { processAuthVars } from "./library/processAuthVars.ts";
+import type { UUID } from "./library/cbor/uuid.ts";
 import {
 	Action,
 	type ActionResult,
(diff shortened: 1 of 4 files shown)


kearfy (approved):

kearfy:
Great find @tai-kun, and thank you for adding some tests around live queries! Hadn't found the time to add them myself but this is great 😃

Issue #242, opened 2024-04-21
Feature: RecordId class should error if passing a RecordId as parameter


vitorTheDev:
### Is your feature request related to a problem?

When upgranding from surreal.js 0.x.x to 1.x.x, it's common to cast string parameters as RecordId class. However, some of those values already come as a RecordId, causing mistakes in casting the recordId twice e.g.
````js
new RecordId('table', new RecordId('table', 'id'))
``` 
This, instead of proper errors, causes weird behavior in surrealdb.

### Describe the solution

The class creation should fail if the id or table parameters are `instanceof RecordId` with a proper error.

### Alternative methods

OR maybe it should parse it correctly.

### SurrealDB version

1.4.2 for windows on x86_64

### SurrealDB.js version

1.0.0-beta.4

### Contact Details

netd777@gmail.com

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

macjuul:
This should no longer be accepted by TypeScript and will result in a clear runtime error in the 2.0 SDK.

Issue #241, opened 2024-04-21
Feature: Implement toString and fromString for RecordId class


vitorTheDev:
### Is your feature request related to a problem?

RecordId usage is pretty common in surrealdb, and casting from/to a string right now is not easy because toString gives `[object Object]`, and to cast it from a string you need to split(':') table and id parts.

### Describe the solution

It would be good if we could just do `id.toString()` because it's the default for JS and will be called automatically when you pass it as a string.
In case of fromString, it would be good if we had a class method for casting a string with both parts as a record id e.g.
```js
RecordId.fromString('table:id')
// or
new RecordId('table:id')
```

### Alternative methods

NONE

### SurrealDB version

1.4.2 for windows 11 x64

### SurrealDB.js version

1.0.0-beta.4

### Contact Details

netd777@gmail.com

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

vitorTheDev:
Just realized you can `new RecordId('table', 'table:id')` without having to split(':') the id parameter... Is this a correct thing to do? Or should I wait for a proper fromString implementation?

vitorTheDev:
Instead of this solution I went for another route: changing the entire app to use only the `.id` part of RecordId. It has solved my problems, but I still think `.toString()` should be more readable so we could console.log it.

PR #244, opened 2024-04-24
feature: add optional `flatMode` flag to flatten surrealdb responses to plain js objects

oskar-gmerek:
## What is the motivation?

Switching from `JSON` to `CBOR` and the subsequent introduction of classes such as `RecordId` significantly lowers the DX in its current form, for example, when working with frameworks like SvelteKit. SvelteKit only supports data transmission between the server and the client in JSON format, and attempting to send the current response from SurrealDB will result in an error Error: Data returned from 'load' while rendering / is not serializable: Cannot stringify arbitrary non-POJOs.

The `toJSON()` method available in the RecordId class solves this problem, but it does not improve DX because before any data can be sent, it must first be destructured and the method used in all occurrences of `RecordId`.

## What does this change do?

This change introduces an optional `flatMode` flag in the constructor of the `Surreal` class. When `flatMode: true`, all responses from SurrealDB that may contain RecordId are recursively flattened into a JavaScript object format via the flatten() function.

Creating new instances:
Regular: `new Surreal()`
flatMode: `new Surreal({ flatMode: true })`

Example Outputs:
Regular: 
```js
{
   id: RecordId {
      tb: "person", 
      id: 1
   },
   firstname: "John",
   lastname: "Doe",
   age: 30,
},
```
flatMode: 
```js
{
   id: {
      tb: "person", 
      id: 1
   },
   firstname: "John",
   lastname: "Doe",
   age: 30,
},
```

I added tests based on existing ones, each test includes the suffix ` - flatMode`. I also compared the performance between the `query` and `query - flatMode` tests, which show a performance difference of less than `1ms`.

#247

Related to, but this is partial or alternative solution for #241 #242 

Also I saw related messages on Discord (but this is partial or alternative solution):
Zrzut ekranu 2024-04-24 o 11 44 02

Zrzut ekranu 2024-04-24 o 11 31 15

Zrzut ekranu 2024-04-28 o 19 33 15

## To do
- [ ] add support for live queries
- [ ] add tests for live queries

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,0 +1,14 @@
+class Config {
+	private _flatMode: boolean = false;
+
+	get flatMode(): boolean {
+		return this._flatMode;
+	}
+
+	set flatMode(value: boolean) {
+		this._flatMode = value;
+	}
+}
+
+const config = new Config();
+export default config;
--- a/src/library/flatten.ts
+++ b/src/library/flatten.ts
@@ -1,0 +1,27 @@
+import { RecordId } from "./cbor/recordid.ts";
+
+/**
+ * Recursively converts CBOR class instances into a JavaScript object format.
+ *
+ * @param {T} obj - The object or array to be flattened to a JavaScript object.
+ * @return {T} The flattened JavaScript object or array.
+ */
+export function flatten<T>(obj: T): T {
+	if (Array.isArray(obj)) {
+		return obj.map(flatten) as unknown as T;
+	} else if (obj !== null && typeof obj === "object") {
+		if (obj instanceof RecordId) {
+			return obj.toJSON() as unknown as T;
+		}
+
+		const flat = {} as T;
+		for (const key in obj) {
+			// deno-lint-ignore no-explicit-any
+			(flat as any)[key] = flatten((obj as any)[key]);
+		}
+
+		return flat;
+	}
+
+	return obj;
+}
--- a/src/surreal.ts
+++ b/src/surreal.ts
@@ -30,6 +30,8 @@ import {
 	TransformAuth,
 } from "./types.ts";
 import { ConnectionStatus } from "./library/engine.ts";
+import config from "./config.ts";
+import { flatten } from "./library/flatten.ts";
 
 type Engines = Record<string, new (emitter: Emitter<EngineEvents>) => Engine>;
 type R = Prettify<Record<string, unknown>>;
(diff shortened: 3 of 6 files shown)


abrisene:
+1 - I have abstraction around instantiating a surreal client in 0.11, but the changes to the ids makes migration more complicated than I'd like.

It would be nice to have this option, which would make it easier to migrate more gradually.

Ideally I'd want to have the option to flatten on the query methods themselves since cbor makes it much easier to use complex ids, but changing the signature of each of the query methods seems like a bad idea.

kearfy:
Hey @oskar-gmerek! Thanks for working on this PR. I don't think we should support global config options like this, we could however improve the toJSON methods in such a way that you can simply do
```js
await db.select('table-name').then(r => JSON.stringify(r))
```
And get roughly the same result back as the previous versions of the library. 
It sounds quite trivial to construct a string RecordID outside the rust library, but it's actually not. Take tb: `table:name` and id `something-else` as an example:
```
"table:name:something-else"
```
You now end up with an invalid record id, as the two ident parts are not escaped. Now let's say that you not send this to SurrealDB, but you would have `RecordId.fromString("table:name:something-else")`, it would also result in something else where tb is `table` and the id part is `name:something-else`.

What I'm boiling down to is that, yes, we can do this, however it would add the cost of using regex which at scale can massively degrade performance. Ideally implementors would adjust to the new library and take advantage of it's understanding of datatypes native to SurrealQL.

I'll close this PR, but I'll see what I can do to add in a JSON compatibility mode. It will however carry a price tag on performance, and you won't be able to interchange some of this values back to the server without manually converting them back to these new representations of native SurrealQL datatypes, which includes RecordId, UUID, Datetime, Duration, Decimal, BigInt, Table and Geometries.

kearfy:
Hey all! As you can see, I just merged a PR which allows you to get a JSON-like representation of the responses the way that SurrealDB would also display JSON-like responses.

Here's an example of what you'll be able to do in the next beta:
```js
// Use the selection, creation and altering methods with string record ids
const res = await db.select(new StringRecordId("person:john"));

// You will get back a response with native values
// However will `toJSON()` or `JSON.stringify()` you will get back a JSON representation of the result, the way that SurrealDB would also display it
const jsonlike = res.toJSON();
const jsonstring = JSON.stringify(res);

// You can also easily do this in one go
await db
    .select(new StringRecordId("person:john"))
    .then(res => res.toJSON());
```

PR #249, opened 2024-04-28
bugfix: display proper error message when surrealdb responds with an error

SamuelScheit:
## What is the motivation?

I was trying to insert a lot of objects and got a the error message "CBORError: Remaining bytes", which I first couldn't understand.
I've traced it back to the engine.ts file and the decodeCbor function and fixed the error message by displaying the response string, which has more meaningful information about the error.

image

## What does this change do?

I've improved the error message by catching any errors thrown by the decodeCbor function and displaying the buffer that is trying to be decoded as an error string.

I've setup surrealdb.js with the patch and tried to insert 100k objects with .insert().
The operation fails with the following error `Failed to buffer the request body: length limit exceeded` instead of `CBORError: Remaining bytes`


kearfy:
Hey @SamuelScheit, thank you for this! There are two issues with the current implementation:
- You throw a generic Error. In this library we only throw errors extending a base `SurrealDbError` class to make it easier for implementors to catch where errors come from.
- `Buffer` is not available in all major runtimes, taking browsers as an example.

I'll see how I can resolve these issues and push those fixes to your branch, unless you'd like to pick them up 😃

SamuelScheit:
@kearfy thank you, you can contribute to my branch 👍

kearfy (approved):

kearfy:
I had to change it up a bit but HTTP Connection errors should now be properly handled. Thanks for opening this!

rvdende:
@kearfy will you bump the version and release a npm version perhaps?

PR #250, opened 2024-04-29
Allow bigint in record ids

### LINKED ISSUE #240: Bug: Missing typing on `.query` method
### Describe the bug

The generic type `T` is not used in the return of the `.query` function:
```
/**
 * Runs a set of SurrealQL statements against the database.
 * @param query - Specifies the SurrealQL statements.
 * @param bindings - Assigns variables which can be used in the query.
 */
async query<T extends unknown[]>(
	query: string | PreparedQuery,
	bindings?: Record<string, unknown>,
) {
	const raw = await this.query_raw<T>(query, bindings);
	return raw.map(({ status, result }) => {
		if (status == "ERR") throw new ResponseError(result);
		return result;
	});
}
```
This makes it so that i get no types on my result.

### Steps to reproduce

Use the latest current version of surrealdb.js and try to specify a type for the `.query` function.

### Expected behaviour

Exptected the result from the function to be of the specified generic type.

### SurrealDB version

surreal 1.4.2 for windows

### SurrealDB.js version

1.0.0-beta.5

### Contact Details

olavnon@gmail.com

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

macjuul:

## What is the motivation?

cbor-redux can decode/encode bigint, however the zod union currently does not list bigint

## What does this change do?

Adds bigint as valid record id value type

Added unit test

If this pull request is related to any other pull request or issue, or resolves any issues - then link all related pull requests and issues here.

- https://github.com/surrealdb/surrealist/issues/241
- Also fixed #240 as it was necessary for the unit test


rluvaton:
FYI, you have a bug here as JSON.stringify does not support big int

See more:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#use_within_json

PR #272, opened 2024-05-30
Migrate to JSR + minor updates

sachaw:

## What is the motivation?

The Deno team has launched [JSR](https://jsr.io/)
You can read about it here: https://deno.com/blog/jsr-is-not-another-package-manager
It would be great to have this package published there

## What does this change do?

This PR changes the following:

1. Updates `deno.json` to use the new fields: `name`, `version`, `description`, replacing `project.json`
2. Uses `deno.json`'s import map to specify package versions, using packages published to JSR where possible
3. Replace `semver` with Deno's `@std/semver`
4. Added as many return type annotations as I could, JSR want's all public return types to be explicitly typed. Right now I have left those returning Zed functions as they are very verbose.
5. Added vscode defaults for default formatter and format on save
6. Added export map to `deno.json` to export `mod.ts`
7. Updated a few CI action versions & added a JSR publish step
8. Add Pull Request CI workflow to check formatting and run tests

What's required if this PR is accepted is for a JSR [JSR](https://jsr.io/) account and namespace to be created, I have specified `@surrealdb` and to register the package name `surrealdb`. Then under the publish settings, specify this github repo to enable automatic provenance and publishing

Thanks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Setup Deno
         uses: denoland/setup-deno@v1
         with:
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Setup deno
         uses: denoland/setup-deno@v1
         with:
@@ -41,7 +41,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Setup deno
         uses: denoland/setup-deno@v1
         with:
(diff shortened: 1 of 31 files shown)


kearfy:
Hey @sachaw, wow! This is an amazing PR, thank you so much! 

One note I have is that I have been using NPM packages with the reason that it's more likely that people already have X npm package installed compared to a Deno or JSR package, which would save a bit on bundle size. For that reason, I would personally opt to stick with the NPM semver package that was previously used. I'm happy to hear your thoughts however!

sachaw:
Thanks for that, Whilst I do agree with not blowing out the bundle size. The `semver` package currently published to npm is a very old, and outdated cjs package, with no native typescript types bundled. While the `@std/semver` package may add a few kb to a project, the increase in quality and adaptability is worth it IMO.

kearfy:
Alright, that's cool! I haven't forgotten your PR by the way, I simply need to find some time to have a proper look and merge it in. This is great!

kearfy:
Hey @sachaw, I really appreciate the effort you put into this project! We were unfortunately planning to change from Deno to Bun for a little while already, which also covers publishing to JSR. 

While I have to close your PR, I'm happy to share that publishing to JSR is added in #272, and all methods are now statically typed as per the JSR guidelines.

Have a nice weekend!

PR #275, opened 2024-06-07
bugfix: Fix the escaping in the built-in ID generation functions

tai-kun:
Thank you for the wonderful project.

## What is the motivation?

The `RecordId` class converts built-in generation functions into strings. For example:

```typescript
const rid = new RecordId("tb", "rand()");
console.log(JSON.stringify(rid)); // "tb:⟨rand()⟩"
```

However, it should ideally be like this:

```typescript
const rid = new RecordId("tb", "rand()");
console.log(JSON.stringify(rid)); // "tb:rand()"

// Manually escape built-in generation functions.
const rid = new RecordId("tb", "⟨rand()⟩");
console.log(JSON.stringify(rid)); // "tb:⟨rand()⟩"
```

## What does this change do?

When escaping `RecordIdValue`, we will not escape `rand()`, `ulid()`, or `uuid()`.

For this implementation, the source code I referred to was:
- https://github.com/surrealdb/surrealdb/blob/v1.5.2/core/src/sql/id.rs#L149-L162

I added these test suites to `tests/unit/jsonify.ts` and updated the snapshots.

N/A


kearfy:
Hey @tai-kun! When you pass a string to the RecordId class, it will be treated as such and not as a generate record id. To support such behaviour, we would also need to treat certain strings special, which is exactly what we are trying to move away from. Additionally, generate record ids are currently not supported by the CBOR protocol: 
- [CBOR to SurrealQL](https://github.com/surrealdb/surrealdb/blob/main/core/src/rpc/format/cbor/convert.rs#L191)
- [SurrealQL to CBOR](https://github.com/surrealdb/surrealdb/blob/main/core/src/rpc/format/cbor/convert.rs#L393-L395)

You are able to do this with the `StringRecordId` class however, which you can use to pass a custom string record id to SurrealDB. To safely construct custom string record ids, we could rather consider to expose the [escape_ident](https://github.com/surrealdb/surrealdb.js/blob/main/src/library/cbor/recordid.ts#L51) method. You could then:
```ts
const rid = new StringRecordId(`${escape_ident("tb-name")}:rand()`);
```

I hope that helps!

PR #289, opened 2024-06-27
publish npm package with latest only for non beta releases

rluvaton:
## What is the motivation?

To avoid installing beta when I run `npm i surrealdb.js` in new projects

## What does this change do?
mark `latest` only versions that match the semver standard, and the rest with `beta` 

Manually tested it

Not existing ones

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -16,6 +16,8 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
       - uses: actions/checkout@v3
       - run: deno task build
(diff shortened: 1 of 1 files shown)


rluvaton:
You probably need a better way to manage versions, as there will come a day that you need to publish hotfix version to multiple version with different majors and it will be really hard

kearfy:
Hey @rluvaton, thanks for this PR! We'll be merging #291 first and I'll merge this right after!

kearfy (approved):
Thanks!

PR #291, opened 2024-06-29
Various improvements

### LINKED ISSUE #284: Bug: MissingNamespaceDatabase: There are no namespace and/or database configured.
### Describe the bug

I keep seeing this error in my logs, even though I am still able to run queries just fine:

```
Trace: Unhandled Rejection Promise: Promise {
  <rejected> MissingNamespaceDatabase: There are no namespace and/or database configured.
      at HttpEngine.rpc (/my-server/node_modules/.pnpm/surrealdb.js@1.0.0-beta.9/node_modules/surrealdb.js/src/library/engine.ts:340:10)
      at async Surreal.ping (/my-server/node_modules/.pnpm/surrealdb.js@1.0.0-beta.9/node_modules/surrealdb.js/src/surreal.ts:173:21)
}
    at process.<anonymous> (/my-server/src/server/functions/server/add.process.error.handling.events.ts:18:11)
    at process.emit (node:events:518:28)
    at process.emit (node:domain:488:12)
    at process.emit.sharedData.processEmitHook.installedValue [as emit] (/my-server/node_modules/.pnpm/@cspotcode+source-map-support@0.8.1/node_modules/@cspotcode/source-map-support/source-map-support.js:745:40)
    at emit (node:internal/process/promises:150:20)
    at processPromiseRejections (node:internal/process/promises:284:27)
    at processTicksAndRejections (node:internal/process/task_queues:96:32)
Trace: error unhandledRejection {
  unhandledRejectionReason: MissingNamespaceDatabase: There are no namespace and/or database configured.
      at HttpEngine.rpc (/my-server/node_modules/.pnpm/surrealdb.js@1.0.0-beta.9/node_modules/surrealdb.js/src/library/engine.ts:340:10)
      at async Surreal.ping (/my-server/node_modules/.pnpm/surrealdb.js@1.0.0-beta.9/node_modules/surrealdb.js/src/surreal.ts:173:21),
  unhandledRejectionPromise: '{}',
  level_name: 'error'
}
```

### Steps to reproduce

```
export default class SurrealDBService {
	public static async init(): Promise<SurrealDBService> {
		if (!SurrealDBService._instance || !SurrealDBService._db) {
			this._db = new Surreal();

			await this._db.connect(config.connections.surrealdb.url);

			await this._db.use({
				namespace: config.connections.surrealdb.namespace,
				database: config.connections.surrealdb.database
			});

			this._token = await this._db.signin({
				username: config.connections.surrealdb.username,
				password: config.connections.surrealdb.password
			});
			console.log('surreal', this._token);
		}

		this._instance = new SurrealDBService();
		return this._instance;
	}

	private static _initialized: boolean = false;
	private static _token: string;
	private static _db: Surreal;
	private static _instance: SurrealDBService;
}
```
All values under config.connections.surrealdb are correct, as I am still able to connect and run querries successfully.

This seems to be an error in the ping call.

### Expected behaviour

not to throw an error...

### SurrealDB version

1.5.3 for linux on x86_64

### SurrealDB.js version

1.0.0-beta.9

### Contact Details

stojadinovicp@gmail.com

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kearfy:

## What is the motivation?

We found that the size of the library was getting out of hand and wanted to improve on this. Additionally, with the introduction of JSR, we no longer see the need for this to be a Deno project.

## What does this change do?

- Drops the `decimal.js` library, as it's unpacked size is 250kb.
- Introduces a custom CBOR library, optimised for size, speed and our needs.
- Introduces a custom Duration library, optimised for size, speed and out needs.
- Introduces a new `Uuid` wrapper to offer a nicer interface over the `uuidv7` library.
- No longer depends on `zod`, which was a big contributor to size.
- No longer depends on `semver`, which was a big contributor to size. We now compare versions with `localeCompare`.
- Adjusted the way that engines work. `encodeCbor` and `decodeCbor` are now passed along with the emitter in an `EngineContext` class, to the engine. This ensures that the engine library uses the same instances of classes that the cbor encoding and decoding depend on.
- Now uses BiomeJS for faster and further optimised formatting.
- Now publishes to the JSR registry
- Fixes certain issues with geometries
- Fixes an issue which caused the Pinger to be applied to engines outside of `ws` and `wss`.
- Optimises the `jsonify` method to use custom logic over `JSON.stringify()` and `JSON.parse()`, keeping non-surrealql class values intact, and resulting in 100% to 150% faster.
- Added more tests.
- We now test across SurrealDB 1.4.2, 1.5.2 and 2.0.0-alpha.3.
- PreparedQuery instances now partially encode queries and bindings before they are sent over.

Updated all tests to Bun, and added various new tests. In the future we'd like to expand tests to test on both source and the compiled library, and across Bun, Deno and Node.

Fixes #284
Supersedes #272


src/data/types/duration.ts
```diff
@@ -0,0 +1,164 @@
...
+const second = 1000 * millisecond;
+const minute = 60 * second;
+const hour = 60 * minute;
+const day = 24 * hour;
+const week = 7 * day;
+
+const units = new Map([
+	["ns", nanosecond],
+	["µs", microsecond],
+	["μs", microsecond],
```
rluvaton, line 13: Duplicated

src/engines/http.ts
```diff
@@ -0,0 +1,155 @@
...
+} from "./abstract";
+
+export class HttpEngine extends AbstractEngine {
+	connection: {
+		url?: URL;
+		namespace?: string;
+		database?: string;
+		token?: string;
+		variables: Record<string, unknown>;
+	} = { variables: {} };
```
rluvaton, line 22: For performance I would initialize them with all frequently used to avoid change the shape of the object

src/engines/ws.ts
```diff
@@ -0,0 +1,205 @@
...
+			});
+
+			socket.addEventListener("close", () => {
+				this.setStatus(ConnectionStatus.Disconnected);
+			});
+
+			socket.addEventListener("message", async ({ data }) => {
+				const decoded = this.decodeCbor(
+					data instanceof Blob
+						? await data.arrayBuffer()
```
rluvaton, line 75: What will catch this if it fail? If nothing the process with crash on unhandled rejection

src/engines/ws.ts
```diff
@@ -0,0 +1,205 @@
...
+export class Pinger {
+	private pinger?: ReturnType<typeof setTimeout>;
+	private interval: number;
+
+	constructor(interval = 30000) {
+		this.interval = interval;
+	}
+
+	start(callback: () => void): void {
+		this.pinger = setInterval(callback, this.interval);
```
rluvaton, line 199: I would unref in environments that support it to prevent from keeping the process alive

src/types.ts
```diff
@@ -1,122 +1,86 @@
...
-			vars.ns = namespace;
-			if (database) {
-				vars.db = database;
-			}
+export function convertAuth(params: AnyAuth): Record<string, unknown> {
+	const cloned: Record<string, unknown> = { ...params };
+	const convertString = (a: string, b: string, optional?: boolean) => {
+		if (a in params) {
+			cloned[b] = `${cloned[a]}`;
+			delete cloned[a];
```
rluvaton, line 20: This can be bad for performance due to changing the object shape

src/util/jsonify.ts
```diff
@@ -0,0 +1,76 @@
...
+			input instanceof StringRecordId ||
+			input instanceof RecordId ||
+			input instanceof Geometry ||
+			input instanceof Table
+		) {
+			return input.toJSON() as Jsonify<T>;
+		}
+
+		// We check by prototype, because we do not want to process derivatives of objects and arrays
+		switch (Object.getPrototypeOf(input)) {
```
rluvaton, line 53: What about null prototype?

rluvaton (commented):
Left some comments

src/data/types/duration.ts
kearfy: Try it!
```js
"µs" == "μs"
```
image

src/engines/ws.ts
kearfy: Nothing, there's no way to point this back to in which place the request was originally made

src/data/types/duration.ts
rluvaton: So I would add a comment

src/engines/http.ts
kearfy: As in make them `undefined` by default?

src/util/jsonify.ts
kearfy: The first check in the function checks for null, so that's already filtered out at this point

src/types.ts
kearfy: Would initiating a new object be more performant?

src/engines/ws.ts
kearfy: As in `delete this.pinger` beforehand?

src/engines/http.ts
rluvaton: Yes

src/engines/ws.ts
rluvaton: You should emit an error event in that case

src/types.ts
rluvaton: Not necessarily, it depends on the pattern of the data

kearfy: It basically rewrites: 
- `namespace` -> `ns`
- `database` -> `db`
- `scope` -> `sc`
- `access` -> `ac`

tests/integration/tests/connection.test.ts
```diff
@@ -0,0 +1,24 @@
...
+	});
+
+	test("version check timeout", async () => {
+		const start = new Date();
+		const res = createSurreal({ reachable: false });
+		const end = new Date();
+		const diff = end.getTime() - start.getTime();
+
+		expect(res).rejects.toBeInstanceOf(VersionRetrievalFailure);
+		expect(diff).toBeLessThanOrEqual(5100); // 100ms margin
```
rluvaton, line 22: I would use a constant for the default timeout + 100 for better readability and avoid magic numbers, also to couple the timeout with it, if the timeout decrease, so should this test

tests/integration/tests/live.test.ts
```diff
@@ -0,0 +1,164 @@
...
+		promise,
+		p.then(resolve).finally(() => clearTimeout(timeout)),
+	]);
+}
+
+describe("Live Queries", async () => {
+	const surreal = await createSurreal();
+
+	test("live", async () => {
+		if (surreal.connection?.connection.url?.protocol.startsWith("http")) {
```
rluvaton, line 21: Avoid conditions inside tests, this can lead to less readable code and worse debugging experience, if the test fail, which path did it took?

You should split to 2 and skip inside if it is not relevant

tests/integration/tests/live.test.ts
```diff
@@ -0,0 +1,164 @@
...
+				firstname: "John",
+				lastname: "Doe",
+			});
+			await surreal.update(new RecordId("person", 1), {
+				firstname: "Jane",
+				lastname: "Doe",
+			});
+			await surreal.delete(new RecordId("person", 1));
+
+			await withTimeout(promise, 5e3); // Wait for the DELETE event
```
rluvaton, line 47: This can be flaky, I would have a helper function to wait for event with timeout if failed to get

tests/integration/tests/querying.test.ts
```diff
@@ -0,0 +1,403 @@
...
+	test("single", async () => {
+		const single = await surreal.create<Person, Omit<Person, "id">>(
+			new RecordId("person", 1),
+			{
+				firstname: "John",
+				lastname: "Doe",
+			},
+		);
+
+		expect(single).toMatchObject({
```
rluvaton, line 37: I think to match object ignore additional keys which I don't think is what you should use here

tests/unit/recordid.test.ts
```diff
@@ -0,0 +1,27 @@
...
+		// Bigint
+		expect(new RecordId("table", 9223372036854775807n).toString()).toBe(
+			"table:9223372036854775807",
+		);
+		expect(new RecordId("table", 9223372036854775808n).toString()).toBe(
+			"table:⟨9223372036854775808⟩",
+		);
+
+		// Objects and arrays
+		// expect(new RecordId("table"))
```
rluvaton, line 25: It seems like you left this commented

rluvaton (commented):
Left some more comments

Also, I would redo the PR to try to keep the git history as most of the files marked as deleted instead of moved and updated, this has really high value

src/engines/ws.ts
```diff
@@ -0,0 +1,205 @@
...
+					);
+				}
+			});
+		});
+
+		this.ready = ready;
+		return await ready.then(() => {
+			this.socket = socket;
+			this.pinger = new Pinger(30000);
+			this.pinger?.start(() => this.rpc({ method: "ping" }));
```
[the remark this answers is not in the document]
kearfy: Typescript was previously not smart enough to deduce that pinger is not undefined if I recall correctly, will adjust

src/engines/ws.ts
```diff
@@ -70,40 +70,48 @@ export class WebsocketEngine extends AbstractEngine {
...
+							: data.buffer.slice(
+									data.byteOffset,
+									data.byteOffset + data.byteLength,
+								),
 					);
+
+					if (
+						typeof decoded === "object" &&
+						decoded != null &&
+						Object.getPrototypeOf(decoded) === Object.prototype
```
rluvaton, line 86: What about null prototype that is often used for better security against prototype pollution

example of how to set it:
```js
{__proto__: null}
```

this is just one example, you can also use `Object.create(null)`

tests/integration/tests/auth.test.ts
```diff
@@ -18,61 +18,75 @@ describe("basic auth", async () => {
...
-				id: 123,
-			});
-
-			expect(typeof signup).toBe("string");
+    		DEFINE TABLE user PERMISSIONS FOR select WHERE id = $auth;
+    		DEFINE SCOPE user
+    			SIGNUP ( CREATE type::thing('user', $id) )
+    			SIGNIN ( SELECT * FROM type::thing('user', $id) );
+    	`);
+	});
```
rluvaton, line 33: I would use before each to make sure the tests are isolated

kearfy: It already uses a single database instance though, and the tests themselves won't affect schema. It should not be an issue this way

src/engines/ws.ts
kearfy: What exactly are you aiming at?

Issue #156, opened 2023-07-26
Bug: NextJS 13 (app router) with Node and Surreal


leolorenzoluis:
### Describe the bug

I'm getting the following and I'm not sure what causes it. I'm using SurrealDB 0.8.3. This is only when it's trying to instantiate and call the db client

```
Module not found: Can't resolve 'bufferutil' in '/Users/meh//node_modules/ws/lib'

Import trace for requested module:
./node_modules/ws/lib/buffer-util.js
./node_modules/ws/lib/receiver.js
./node_modules/ws/wrapper.mjs
./node_modules/unws/src/node.mjs
./node_modules/surrealdb.js/esm/library/WebSocket/node.js
./node_modules/surrealdb.js/esm/library/SurrealSocket.js
./node_modules/surrealdb.js/esm/strategies/websocket.js
./node_modules/surrealdb.js/esm/index.js
./db.ts
./auth.ts
./app/api/auth/[...nextauth]/route.ts

```

### Steps to reproduce

```
export const getDbClient = async () => {
    await db.signin({
        user: 'bla',
        pass: 'blah'
    });
    await db.use({ ns: test', db: 'test' });
    return db;
}
```

### Expected behaviour

It should not throw

### SurrealDB version

Release v1.0.0-beta.9

### SurrealDB.js version

0.8.3

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kearfy:
Hi Leo, 

This is not something we can solve, but there is an easy workaround: https://github.com/vercel/next.js/issues/44273

Simply adjust the config of webpack in your `nextjs.config.js` file to include the following bit (provided a complete config file as example):
```js
/** @type {import('next').NextConfig} */
const nextConfig = {
    webpack: (config) => {
        config.externals.push({
            'utf-8-validate': 'commonjs utf-8-validate',
            bufferutil: 'commonjs bufferutil',
        });
        return config;
    },
};

module.exports = nextConfig;
```

Let me know if this helps!
Micha

leolorenzoluis:
Amazing find i tried searching it didn’t come up. You have better google skills than me! Thank u

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #337, opened 2024-09-04
Remove unnecessary optional dependecies

lkwr:

## What is the motivation?

The current version of `surrealdb.js` uses the optional dependencies `bufferutil` and `utf-8-validate` which are unnecessary.

The two libraries are optional dependencies of the `ws` package. The `ws` package is only required on runtimes that not implement WebSockets (see `isows` implementation). So it is not necessary to include these optional dependencies here in `surrealdb.js`.

The problem is that the package manager always try to install optional dependencies which is useless on runtimes where WebSocket are supported. It even uses `node-gyp` which may not always work perfectly on alternative runtimes like bun.

This change should **not** break anything because "non-WebSocket-Runtimes" are required to install the `ws` package (peer dep in `isows`) which already includes `bufferutil` and `utf-8-validate` as optional dependencies.

## What does this change do?

Remove the optional dependencies `bufferutil` and `utf-8-validate`.

CI

Didn't find anything.


kearfy:
For context, the reason that we include them as optional deps is because compilers would complain about them, which then required extra config. We can exclude them if we document these pitfalls better. @Ekwuno what are your thoughts? Specifically nextjs used to be quite problematic

kearfy:
This was the original issue which led me to include them: https://github.com/surrealdb/surrealdb.js/issues/156

Besides the countless reports in our discord server

lkwr:
@kearfy thanks for the notice, i missed this :)

I tested it myself and can confirm the missing `bufferutil` and `utf-8-validate` error on nextjs 13 (with node `18.20.4`/`22.8.0`).

But by using the latest version of nextjs (which is 14) the error is gone (without `bufferutil`/ `utf-8-validate` installed!)

---

It seems to be a nextjs 13 issue. Take a look at the code which causes the `bufferutil` error:

`node_modules/ws/lib/buffer-util.js`
```javascript
// ...

/* istanbul ignore else  */
if (!process.env.WS_NO_BUFFER_UTIL) {
  try {
    const bufferUtil = require('bufferutil');

    module.exports.mask = function (source, mask, output, offset, length) {
      if (length < 48) _mask(source, mask, output, offset, length);
      else bufferUtil.mask(source, mask, output, offset, length);
    };

    module.exports.unmask = function (buffer, mask) {
      if (buffer.length < 32) _unmask(buffer, mask);
      else bufferUtil.unmask(buffer, mask);
    };
  } catch (e) {
    // Continue regardless of the error.
  }
}
```

Thats the only import of `bufferutil` at all. An error should not be thrown because the whole require statement is wrapped into a try-catch block. So its intended that this may fail. (same goes for `utf-8-validate`).

It seems more like nextjs is ignoring the try-catch block (at least in v13). Maybe something like a pre analysis of required modules.

---

IMHO its not `surrealdb.js`'s job to prevent this kind of error, especially if nextjs 14 solved this issue already for me. Developers who encounter this error should install the dependencies themselves.

But I think there should be a note in the README/docs that non websocket runtimes like `node <22` need the additional `ws` package (but this is not directly related to this).


What are your thoughts?

kearfy (approved):

kearfy:
Alright, yes, I agree!

PR #344, opened 2024-09-25
Misc improvements

macjuul:
- Refactored file names and imports
- Enforce file name format
- Initial work on reconnect support


ntorrey:
Yay for reconnect support! 🎉

phisch:
@macjuul We just ran into an issue of our backend losing the connection to the database, and the backend had no way to even tell us about that. Anything using queries was just stuck indefinitely. We'd be super stoked about reconnect support. But is there anything we can do in the meantime until this gets merged? Even just being able to have a timeout that would let us know the connection is lost would be a big improvement.

macjuul:
> @macjuul We just ran into an issue of our backend losing the connection to the database, and the backend had no way to even tell us about that. Anything using queries was just stuck indefinitely. We'd be super stoked about reconnect support. But is there anything we can do in the meantime until this gets merged? Even just being able to have a timeout that would let us know the connection is lost would be a big improvement.

We are still debating on the internal implementation of a reconnect system, however in Surrealist I implemented this manually by subscribing to disconnect events and instantiating a new connection. Until official support is available this is likely the best approach

ntorrey:
@macjuul Ooh, I never knew there was an `.emitter` to hook in to! I don't know if you're able, but could you point me in the right direction where you implement the manual reconnection in surrealist? I'm banging my head against the wall trying to figure it out. Or might you be able to share your implementation @phisch ? Thanks for the help and sorry for the bother!

phisch:
@ntorrey I can't share my implementation, but something along the lines of this should get you started:

```ts
const connection: Surreal = new Surreal();

async function connect() {
	await connection.connect(env.SURREALDB_URL, {
		namespace: env.SURREALDB_NAMESPACE,
		database: env.SURREALDB_DATABASE_NAME,
	});
}

connection.emitter.subscribe(ConnectionStatus.Connected, async () => {
	console.log("Connected to database!");
});

connection.emitter.subscribe(ConnectionStatus.Disconnected, async () => {
	connect();
});

connect();
```

I also handle browser online/offline state changes, have a connect retry-strategy with backoff in there and some more. But this should be a working start.

ntorrey:
@phisch Thank you for the response! I started tinkering and have more questions - I guess this isn't really the place for this, so I'll just ask on discord. Thanks again though!

macjuul:
Reconnect functionality has been superseded by #385

PR #358, opened 2024-10-11
Value deep equality function

macjuul:

## What is the motivation?

There is currently no convenient way to compare SurrealQL values, such as UUIDs or record ids.

## What does this change do?

Adds a new `equals` function to each SurrealQL value type, as well as an exported `equals` function which performs deep equality on two values.

Added a unit test

If this pull request is related to any other pull request or issue, or resolves any issues - then link all related pull requests and issues here.

### DIFF
--- a/src/data/types/future.ts
+++ b/src/data/types/future.ts
@@ -7,7 +7,6 @@ export class Future extends Value {
 
 	equals(other: unknown): boolean {
 		if (!(other instanceof Future)) return false;
-		// TODO ignore whitespace
 		return this.inner === other.inner;
 	}
 
--- a/src/data/types/future.ts
+++ b/src/data/types/future.ts
@@ -1,5 +1,15 @@
-export class Future {
-	constructor(readonly inner: string) {}
+import { Value } from "../value";
+
+export class Future extends Value {
+	constructor(readonly inner: string) {
+		super();
+	}
+
+	equals(other: unknown): boolean {
+		if (!(other instanceof Future)) return false;
+		// TODO ignore whitespace
+		return this.inner === other.inner;
+	}
 
 	toJSON(): string {
 		return this.toString();
--- a/src/data/types/range.ts
+++ b/src/data/types/range.ts
@@ -1,5 +1,6 @@
 import { Tagged } from "../../cbor";
 import { SurrealDbError } from "../../errors";
+import { equals } from "../../util/equals";
 import { toSurrealqlString } from "../../util/to-surrealql-string";
 import { TAG_BOUND_EXCLUDED, TAG_BOUND_INCLUDED } from "../cbor";
 import { Value } from "../value";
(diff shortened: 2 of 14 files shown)


src/data/types/future.ts
```diff
@@ -1,5 +1,15 @@
...
+import { Value } from "../value";
+
+export class Future extends Value {
+	constructor(readonly inner: string) {
+		super();
+	}
+
+	equals(other: unknown): boolean {
+		if (!(other instanceof Future)) return false;
+		// TODO ignore whitespace
```
kearfy, line 10: that is dangerous haha, strings need to be unaffected

src/data/value.ts
```diff
@@ -0,0 +1,11 @@
+export abstract class Value {
+	/**
+	 * Compare equality with another value.
+	 */
+	abstract equals(other: unknown): boolean;
+
+	/**
+	 * Convert this value to a serializable string
+	 */
+	abstract toJSON(): unknown;
```
kearfy, line 10: Can we add toString too?

src/util/equals.ts
```diff
@@ -0,0 +1,38 @@
...
+	if (x instanceof Value && y instanceof Value) {
+		return x.equals(y);
+	}
+	if (
+		typeof x !== "object" ||
+		x === null ||
+		typeof y !== "object" ||
+		y === null
+	) {
+		return false;
```
kearfy, line 28: This includes all basic values right? strings, null, undefined, numbers, etc
```suggestion
		return x === y;
```

src/data/types/recordid.ts
```diff
@@ -35,16 +44,21 @@ export class RecordId<Tb extends string = string> {
...
+		super();
 		if (typeof rid !== "string")
 			throw new SurrealDbError("String Record ID must be a string");
-
 		this.rid = rid;
 	}
 
+	equals(other: unknown): boolean {
+		if (!(other instanceof StringRecordId)) return false;
+		return this.rid === other.rid;
```
kearfy, line 59: Should we then also somehow ignore whitespace here?

src/data/types/range.ts
```diff
@@ -1,19 +1,33 @@
...
+		super();
+	}
+
+	equals(other: unknown): boolean {
+		if (!(other instanceof Range)) return false;
+		if (this.beg?.constructor !== other.beg?.constructor) return false;
+		if (this.end?.constructor !== other.end?.constructor) return false;
+		return (
+			equals(this.beg?.value, other.beg?.value) &&
+			equals(this.end?.value, other.end?.value)
```
kearfy, line 28: Don't the bound values also need to implement `Value` and `.equals()` for this to work?

src/data/types/future.ts
macjuul: It was more so with the thought of ql-wasm comparison, however since I don't see that being viable I'll just remove the comment for now

src/data/types/range.ts
macjuul: It delegates back to the global equals function which also compares primitives, arrays, and objects, so no!

src/data/types/recordid.ts
macjuul: Since string record ids are only used for encoding and not decoding, I would leave the responsibility for whitespace handling to the user and not perform any magic here

src/util/equals.ts
macjuul: Primitives, null, and undefined are already handled by the `Object.is` preceding it

src/data/types/recordid.ts
kearfy: Yep agree, was just trying to make a point about the whitespace haha

kearfy (approved):

Issue #301, opened 2024-07-04
Feature: Convert RecordId from Class to JSON-object while maintaining JSON-structure


OTheNonE:
### Is your feature request related to a problem?

I have a problem regarding using surrealdb.js on the server and having to send data from the server to the client. I want to send the RecordId of an record to the client in a JSON format instead of a `tb:id`-string. Though when trying to do this, i stumble upon the following problems:

- It is not possible to send classes from server to client.
- When using `jsonify`, the RecordId is converted into a string of the form `tb:id`.

Therefore, i have no simple way of returning data from the server to the client while maintaining the JSON-object in the recordId.

### Describe the solution

It would be nice to have a utility function or a method implemented on the class that could convert the recordId class into JSON and maintain the JSON-structure:
```javascript
const rid = new RecordId("recording", { city: "London", data: 123 });
const rid_json = rid.toJSONObject()
rid_json.city // "London"

// Send the rid_json:
fetch("https://example.com/abc", {
  method: "POST",
  headers: {
    "Content-Type": "application/json",
  },
  body: JSON.stringify(rid_json),
})
```

### Alternative methods

Write my own utility function, or also store the data elsewhere than in the recordId.

### SurrealDB version

surreal 1.5.3

### SurrealDB.js version

1.0.0-beta-12

### Contact Details

olavnon@gmail.com

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kearfy:
Heya! You should be able to simply obtain the id part by selecting the `id` property from the `RecordId`. The `jsonify` utility simply outputs the JSON-like value that SurrealDB would

akkie:
I also have this problem with SvelteKit. Only POJOs can be returned from load functions. Would be great if the SDK would have a function that converts the record ID to a POJO. In the meantime here is a workaround:

```ts
function convertRecordIds<T>(obj: T): T {
  if (obj instanceof RecordId) {
    return { tb: obj.tb, id: obj.id } as T;
  } else if (typeof obj === 'object' && obj !== null) {
    return Object.fromEntries(Object.entries(obj).map(([key, value]) => [key, convertRecordIds(value)])) as T;
  } else {
    return obj;
  }
}
```

akkie:
[...]

PR #366, opened 2024-11-01
fix: RecordId's toJSON returns object, not string

ThatOneCalculator:

## What is the motivation?

When passing data from a backend to frontend, `toJSON()` is typically called and is expected to return an object, but instead it returns `${tb}:${id}`. This is not expected from a `toJSON()` function.

## What does this change do?

`toJSON()` now returns `{ tb: string, id: RecordIdValue }`

We tested this in production :^)

#301 

### DIFF
--- a/src/data/types/recordid.ts
+++ b/src/data/types/recordid.ts
@@ -33,8 +33,11 @@ export class RecordId<Tb extends string = string> extends Value {
 		return this.tb === other.tb && equals(this.id, other.id);
 	}
 
-	toJSON(): string {
-		return this.toString();
+	toJSON(): { tb: string; id: RecordIdValue } {
+		return {
+			tb: escape_ident(this.tb),
+			id: this.id,
+		};
 	}
 
 	toString(): string {


kearfy:
Hey @ThatOneCalculator, this behaviour is actually expected. SurrealDB's JSON representation for Record IDs is stringified, a behaviour which the JS SDK replicates too. If you require different behaviour, the jsonify method, recursively modifying values in a type-safe manner, is a good example to copy and alter as you wish.

ThatOneCalculator:
But `toJSON` in JavaScript is a reserved function for returning an object in a JSON-consumable manner, much in the same way the `toString` is reserved for that same purpose for strings. `key:id` is not valid json, and when we  frameworks take a RecordId as something to return, they expect it to turn into a keyed object, not a string.

lkwr:
@ThatOneCalculator a string is valid JSON value, so its totally fine to return a string in an `toJSON` method.
See for example the `Date.prototype.toJSON` method, which also returns a string. ([source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toJSON))

ThatOneCalculator:
And I think this goes back to the actual problem I was seeing: when `toJSON` returns something like `user:bob`, when sending RecordIds from our backend to our frontend, we have to have `user.id.replace("user:", "");` either in the backend or frontend at one point or another to actually get the value of the ID, which gets messy fast, however, when RecordIds are represented as `{ tb: string, id: string | number }`, it's just `user.id.id`.

PR #387, opened 2024-12-08
Allow custom CBOR encoding/decoding replacer

lkwr:

## What is the motivation?

I'm currently building an ORM/ODM for SurrealDB in TypeScript, which I want to open source soon (hopefully this year). For the underlying connection to the database, I use `surrealdb.js`. But types like `RecordId` or `Table` are reimplemented so I can provide a much deeper integration in the whole library.

And there comes the problem: Currently before passing the query to `surrealdb.js` I have a convert step which converts all "custom types" to `surrealdb.js` equivalents. It would be faster and more flexible when I'm able to overwrite the default CBOR encoding/decoding behaviour.

## What does this change do?

So this PR does exactly that, it provides a way to overwrite the default CBOR encoding/decoding replacer. I also export the default replacer, so its much easier to add custom types on top of the default ones.

What are your thoughts on this?

--

--


raymonddaikon:
This would be nice as I'm currently patching the decoding methods to return only the id section of RecordId as a string as well as Date().toISOString() so I can pass ids around my application without normalizing RecordIds every step of the way (this is primarly an issue with deeply nested subqueries).

Issue #389, opened 2024-12-13
Bug: Missing escape characters for long, numeric string identifiers


lylejohnson:
### Describe the bug

A `RecordId` with a long string of numbers as its identifier is not always jsonified properly. In some cases the long numeric string is not properly escaped.

### Steps to reproduce

Here's a simple example to demonstrate the problem:
```
import { jsonify, RecordId } from 'surrealdb';

jsonify(new RecordId('foo', '125813199042576601589342522460260755'));
=> foo:125813199042576601589342522460260755
```

### Expected behaviour

I would expect the result to be `foo:⟨125813199042576601589342522460260755⟩`, for consistency with how SurrealDB treats them:
```
-- string identifier consisting of a long string of numbers
CREATE foo:`125813199042576601589342522460260755`;

[
  {
    id: foo:⟨125813199042576601589342522460260755⟩
  }
]
```

### SurrealDB version

2.1.2

### JavaScript SDK version

1.1.0

### Contact Details

lyle@lylejohnson.name

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #393, opened 2025-02-01
Update recordid.ts

CLoaKY233:
# Fix: Proper Escaping of Long Numeric String Record IDs

## Issue Description
ISSUE --->      https://github.com/surrealdb/surrealdb.js/issues/389#issue-2738662951
When handling RecordId instances with long numeric strings as identifiers (15 or more digits), the current implementation doesn't properly escape these values. This causes inconsistency with SurrealDB's native behavior, which wraps such identifiers in angle brackets (⟨⟩).

## Current Behavior
```typescript
new RecordId('foo', '125813199042576601589342522460260755').toString()
// Returns: foo:125813199042576601589342522460260755
```

## Expected Behavior
```typescript
new RecordId('foo', '125813199042576601589342522460260755').toString()
// Should return: foo:⟨125813199042576601589342522460260755⟩
```

## Changes Made
1. Modified `escapeIdPart` function to detect and properly escape long numeric string IDs
2. Updated `StringRecordId` constructor to handle long numeric string IDs consistently
3. Added specific handling for IDs matching the pattern `/^\d{15,}$/`

## Technical Details
- Added check for numeric strings of 15 or more digits
- Implemented wrapping with angle brackets (⟨⟩) for matching IDs
- Preserved existing behavior for all other ID types
- Maintains backward compatibility with existing implementations

## Testing
Tested with various ID types including:
- Regular string IDs
- Short numeric strings
- Long numeric strings
- UUIDs
- Complex objects and arrays

## Examples

```typescript
// Long numeric string ID
new RecordId('table', '125813199042576601589342522460260755')
// => table:⟨125813199042576601589342522460260755⟩

// Regular string ID (unchanged behavior)
new RecordId('table', 'regular-id')
// => table:regular-id

// Numeric ID (unchanged behavior)
new RecordId('table', 123)
// => table:123

### DIFF
--- a/src/data/types/recordid.ts
+++ b/src/data/types/recordid.ts
@@ -68,15 +68,16 @@ export class StringRecordId extends Value {
 			throw new SurrealDbError("String Record ID must be a string");
 		}
 	}
+
 	equals(other: unknown): boolean {
 		if (!(other instanceof StringRecordId)) return false;
 		return this.rid === other.rid;
 	}
-	
+
 	toJSON(): string {
 		return this.rid;
 	}
-	
+
 	toString(): string {
 		return this.rid;
 	}
@@ -98,21 +99,11 @@ export function isValidIdPart(v: unknown): v is RecordIdValue {
 }
 
 export function escapeIdPart(id: RecordIdValue): string {
-    if (id instanceof Uuid) {
-        return `u"${id}"`;
-    }
-    
-    if (typeof id === "string" && /^\d{15,}$/.test(id)) {
-        return `⟨${id}⟩`;
-    }
-    
-    if (typeof id === "string") {
-        return escapeIdent(id);
-    }
-    
-    if (typeof id === "bigint" || typeof id === "number") {
-        return escapeNumber(id);
-    }
-    
-    return toSurrealqlString(id);
+	return id instanceof Uuid
+		? `u"${id}"`
+		: typeof id === "string"
+			? escapeIdent(id)
+			: typeof id === "bigint" || typeof id === "number"
+				? escapeNumber(id)
+				: toSurrealqlString(id);
 }
--- a/src/data/types/recordid.ts
+++ b/src/data/types/recordid.ts
@@ -51,28 +51,23 @@ export class RecordId<Tb extends string = string> extends Value {
  * A SurrealQL string-represented record ID value.
  */
 export class StringRecordId extends Value {
-    public readonly rid: string;
+	public readonly rid: string;
 
-    constructor(rid: string | StringRecordId | RecordId) {
-        super();
+	constructor(rid: string | StringRecordId | RecordId) {
+		super();
 
-        if (rid instanceof StringRecordId) {
-            this.rid = rid.rid;
-        } else if (rid instanceof RecordId) {
-            this.rid = rid.toString();
-        } else if (typeof rid === "string") {
-            // Handle long numeric string IDs in the constructor
-            const [table, id] = rid.split(':');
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/data/types/recordid.ts [hunk 1]
```diff
@@ -51,33 +51,37 @@ export class RecordId<Tb extends string = string> extends Value {
...
+            const [table, id] = rid.split(':');
+            if (id && /^\d{15,}$/.test(id)) {
+                this.rid = `${table}:⟨${id}⟩`;
+            } else {
+                this.rid = rid;
+            }
+        } else {
+            throw new SurrealDbError("String Record ID must be a string");
+        }
+    }
```
kearfy, line 74: We assume that the user passes a correct record id here. We purposefully do not parse record ids in the JS SDK, as they can contain a huge subset of surrealql values, making them vastly complex to parse outside the surrealql parser. Could you revert this part of your PR?

src/data/types/recordid.ts
```diff
@@ -99,11 +103,21 @@ export function isValidIdPart(v: unknown): v is RecordIdValue {
...
-			: typeof id === "bigint" || typeof id === "number"
-				? escapeNumber(id)
-				: toSurrealqlString(id);
+    if (id instanceof Uuid) {
+        return `u"${id}"`;
+    }
+    
+    if (typeof id === "string" && /^\d{15,}$/.test(id)) {
+        return `⟨${id}⟩`; // Wrap long numeric strings in angle brackets
+    }
```
kearfy: Every possible string which can be parsed as a number needs to be escaped no matter the length. The escapeIdent method takes care of that, however it's logic to check if a string is only numbers depends on `parseInt`. With longer numbers that probably fails with `NaN` cause the number-like string not to be escaped. Could you instead for the `isOnlyNumbers` method used by `escapeIdent` to use this regex, or perhaps a simple for-loop? Not sure which one will be more performant

kearfy:
Thank you for your PR @CLoaKY233! Left two comments 🙏

src/data/types/recordid.ts
kearfy: Did a little check, a regex will indeed be the fastest 🙏

CLoaKY233:
Following the review feedback, I've simplified the fix to focus on the root cause:

1. Changed `isOnlyNumbers` function to use regex instead of `parseInt`:
```typescript
function isOnlyNumbers(str: string): boolean {
    return /^\d+$/.test(str.replace(/_/g, ''));
}
```

Key improvements:
- Replaced `parseInt`-based check with regex pattern `/^\d+$/`
- Added global flag to underscore replacement (`/_/g`) to handle multiple underscores
- Removed unnecessary string conversion and comparison
- More reliable detection of numeric strings of any length

src/util/escape.ts
```diff
@@ -51,11 +51,10 @@ export function escape_ident(str: string): string {
  * @returns Optionally escaped number
  */
 export function escapeNumber(num: number | bigint): string {
-	return num <= MAX_i64 ? num.toString() : `⟨${num}⟩`;
+    return num <= MAX_i64 ? num.toString() : `⟨${num}⟩`;
 }
 
+// THIS IS THE ONLY CHANGE
```
kearfy, line 57:
```suggestion
```

kearfy:
Awesome @CLoaKY233, thank you so much! One more favor, could you run `bun quality:apply:unsafe` to fix the "Code Quality" test? Thank you 🙏

kearfy (approved):

kearfy:
Thanks for bearing with me @CLoaKY233! Merged it!

Issue #409, opened 2025-04-01
Feature: Transaction support


knackstedt:
### Is your feature request related to a problem?

I need to run database queries and provide some context variables to table Events. A solution I have identified is having SDK support for Transactions which I can add variables to.

### Describe the solution

A Transaction class that lets a developer group database calls together into one query that can be executed with variables which can be propagated to the table Event.

In my case, I need to share OTEL span and trace IDs to the events to map traces to updates in the system.

### Alternative methods

Using db.let (which isn't synchronous -- it would easily have consistency issues and pollute other record mutations)
Writing a custom Surreal wrapper that converts each method into a surql query.

### SurrealDB version

n/a

### JavaScript SDK version

1.2.1

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #410, opened 2025-04-01
Proposal: Transaction class to execute actions with context variables.

knackstedt:

Ref: #409 

## What is the motivation?

I needed transaction like support in order to attach some user and OTEL information onto table events. I realized I could create a transaction and have my variables scoped to the transaction. 

## What does this change do?

This change adds a proposed `Transaction` class that can have multiple statements added in order and executed within a SurrealQL Transaction.

This change lacks RecordId support for JSON objects, and it includes several stubs. I created this `transaction` class as it seems to be the most direct way of supplying context variables to the underlying SurrealDB instance without a major API overhaul across both the SDK and the network communication layer.

I did not write tests as this is a proposal / provisional sample.

#409 

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -7,6 +7,7 @@ export * from "./cbor/error";
 export * from "./data";
 export * from "./errors.ts";
 export * from "./types.ts";
+export * from "./transaction.ts";
 export * from "./util/equals.ts";
 export * from "./util/jsonify.ts";
 export * from "./util/version-check.ts";
--- a/src/surreal.ts
+++ b/src/surreal.ts
@@ -202,8 +202,8 @@ export class Surreal {
 
 	/**
 	 * Switch to a specific namespace and database.
-	 * @param database - Switches to a specific namespace.
-	 * @param db - Switches to a specific database.
+	 * @param namespace - Switches to a specific namespace.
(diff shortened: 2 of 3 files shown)


kearfy:
Hey @knackstedt, we're planning on offering client-side transactions in the future. At the moment, we try to keep SDKs as clean as possible, in the sense that they stick as much as possible to only what the RPC protocol offers functionality-wise. Any added functionality can be added as a layer on top of the SDK, in a separate package. I'll therefor be closing this PR at this time.

As a sidenote, the SDK actually offers a `toSurrealqlString()` method, which is a bit more correct over your implementation. If you wish to use this Transaction class in your own project, or to offer it as a separate package on NPM, you could consider using that instead: https://github.com/surrealdb/surrealdb.js/blob/main/src/util/to-surrealql-string.ts#L18

Hope this helps! Feel free to ping me for any further questions (I usually respond faster on Discord, by the way)

knackstedt:
@kearfy Thanks, I figured this would happen. I was primarily trying to propose the concept and how it may work -- as I needed this functionality in my project sooner than the timeline you're able to work on it. It's also very good to learn about that toSurrealqlString method, I figured one might be in the codebase but I didn't search the right keyword to find it :smile: 
Thanks!

Issue #492, opened 2025-11-07
Feature: query accepting CBOR vars


orimay:
### Is your feature request related to a problem?

Currently, if we want to use SurrealDB in a WebWorker, we need to encode vars via CBOR, send it to the WebWorker, parse them into object and feed to `query`. However, the engine already does CBOR encoding/decoding, so moving WebWorker there could greatly improve performance. This will be very noticable on mobile devices, as WASM code runs on the main thread

### Describe the solution

We could feed CBOR-encoded vars directly to `WebWorker` in order to not parse them twice

### Alternative methods

A clumsy workaround is to encode/decode query and its bindings and pass to WebWorker (I currently use this approach), but this makes us use CBOR encoding/decoding twice for the same data for no good reason

### SurrealDB version

2.3.10

### JavaScript SDK version

2.3.4

### Contact Details

dmitrii.a.baranov@gmail.com

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct (repository /surrealdb.js)

- [x] I agree to follow this project's Code of Conduct

PR #493, opened 2025-11-07
Move WASM to WebWorker

orimay:

## What is the motivation?

WASM code executes on the main thread, therefore all the queries are blocking (even though they are async), which is extremely noticable on mobile devices

## What does this change do?

Since CBOR data can be moved from/to WebWorker for free, I simply moved all the DB calls to WebWorker, keeping interface the same way it used to be. I use similar approach locally, but I use CBOR to encode/decode data out of the engine, which adds a slowdown

Sadly, I wasn't able to find a way to build and test it locally, as I am getting `Cannot find module '../wasm/surrealdb' or its corresponding type declarations.ts(2307)` error

https://github.com/surrealdb/surrealdb.js/issues/492


orimay:
Some types and consts are imported from that file (`Request`, `Response`)

macjuul (approved):

PR #519, opened 2025-12-19
Circular dependency when working with bundlers

msanchezdev:

## What is the motivation?

Currently we are trying to export the `base` bundle in the `server` bundle through the package name. This causes a circular dependency when bundlers resolve to the same entrypoint and cache doesn't hit. Noticed while working on a Tanstack Start project where it wasn't possible to import the `surrealdb` package on the server side as it was always resolving to be undefined.

## What does this change do?

Introduce a rolldown plugin to use the correct bundle according to the output format.

Verified a tanstack start project works properly when bundles are build with this setup.

No

### DIFF
--- a/packages/sdk/build.ts
+++ b/packages/sdk/build.ts
@@ -10,7 +10,24 @@ await bundle.write({ format: "cjs", file: "./dist/surrealdb.cjs" });
 // Server-side bundle (Node/Bun/Deno)
 const serverBundle = await rolldown({
     input: "./src/index.server.ts",
-    external: ["node:util", "surrealdb"],
+    external: ["node:util"],
+    plugins: [
+        {
+            name: "use-base-bundle",
+            resolveId(source) {
+                if (source === "surrealdb") {
+                    // @ts-expect-error: outputOptions is not typed in rolldown's plugin API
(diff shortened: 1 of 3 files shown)


msanchezdev:
> @msanchezdev what are the typing changes about exactly?

@kearfy 
Hi, they are actually a shorthand version for the constructor to refer to other RecordId type without having to split their parts
```ts
// Say we have a UserId like:
type UserId = RecordId<"user", [string, { country: string, company: string }]>;
// Makes possible to do:
const user1 = new RecordId<UserId>("user", ["test", { country: "DO", company: "SurrealDB" }]);
// Instead of:
const user2 = new RecordId<UserId['table'], UserId['id']>("user", ["test", { country: "DO", company: "SurrealDB" }]);
```

I wanted to add this quiclky in this PR but if you want I can move it to a different PR. Just wanted to point out that this PR needs to get out ASAP because #503 made wrong implementation of exporting from main 'surealdb' bundle.

I had told Julian before to check it before the release, but I just saw the Product Update for the JS SDK and came to confirm if this PR was merged already, do you think we can get this deployed today as some people may have problem with the latest release?

Issue #585, opened 2026-03-20
Bug: new releases broken in Deno


lenovouser:
### Describe the bug

App crashes because it cannot find `@surrealdb/node-linux-arm64-gnu`

### Steps to reproduce

```
error: Uncaught (in promise) Error: Cannot find native binding. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
    at file:///tmp/deno-compile/node_modules/@surrealdb/node/dist/surrealdb-node.mjs:401:35
Caused by: [
  TypeError: /tmp/1264907040690229177010738459959747053880.node: undefined symbol: _ZTVSt12bad_weak_ptr
      at Object..node (node:module:863:20)
      at Module.load (node:module:680:32)
      at Module._load (node:module:538:12)
      at Module.require (node:module:699:19)
      at require (node:module:874:16)
      at requireNative (file:///tmp/deno-compile-/node_modules/@surrealdb/node/dist/surrealdb-node.mjs:242:11)
      at file:///tmp/deno-compile-/node_modules/@surrealdb/node/dist/surrealdb-node.mjs:387:17,
  Error: Cannot find module '@surrealdb/node-linux-arm64-gnu'
  Require stack:
  - /tmp/deno-compile-/node_modules/@surrealdb/node/dist/surrealdb-node.mjs
      at Module._resolveFilename (node:module:638:15)
      at Module._load (node:module:501:27)
      at Module.require (node:module:699:19)
      at require (node:module:874:16)
      at requireNative (file:///tmp/deno-compile-/node_modules/@surrealdb/node/dist/surrealdb-node.mjs:247:20)
      at file:///tmp/deno-compile-/node_modules/@surrealdb/node/dist/surrealdb-node.mjs:387:17 {
    code: "MODULE_NOT_FOUND",
    requireStack: [
      "/tmp/deno-compile-/node_modules/@surrealdb/node/dist/surrealdb-node.mjs"
    ]
  }
]
```

### Expected behaviour

No crashes.

### SurrealDB version

3.0.3

### JavaScript SDK version

2.0.3

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct (repository /surrealdb.js)

- [x] I agree to follow this project's Code of Conduct

Issue #575, opened 2026-03-11
Bug: Bun native bindings


Odonno:
### Describe the bug

I have this error when using surrealdb in a elctrobun app:

```
Uncaught exception in worker: 84580 |         if (process.env.NAPI_RS_FORCE_WASI)
84581 |           loadErrors.push(err);
84582 |       }
84583 |   }
84584 |   if (!nativeBinding) {
84585 |       throw new Error("Cannot find native binding. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.", { cause: loadErrors });
                    ^
error: Cannot find native binding. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
```

### Steps to reproduce

Add the following code to a electrobun app:

```ts
import { createNodeEngines } from "@surrealdb/node";
import { Surreal } from "surrealdb";

const createDbClient = async () => {
	try {
		const db = new Surreal({
			engines: {
				...createNodeEngines(),
			},
		});

		await db.connect("rocksdb://./data");
		await db.use({ namespace: "main", database: "main" });
	} catch (e) {
		console.error(e);
		throw e;
	}
};

const db = await createDbClient();
```

Note: "@surrealdb/node": "3.0.3"

### Expected behaviour

It should work without any error.

### SurrealDB version

N/A

### JavaScript SDK version

2.0.2

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct (repository /surrealdb.js)

- [x] I agree to follow this project's Code of Conduct

mdrv:
@Odonno Were you perhaps using ARM Linux GNU device? You can try replacing `surrealdb-node.linux-arm64-gnu.node` with [the one from `@mdrv/surrealdb-node`](https://unpkg.com/@mdrv/surrealdb-node@3.0.3/dist/surrealdb-node.linux-arm64-gnu.node) or simply replace `@surrealdb/node` imports with mine (at least until the fix gets merged).

Odonno:
Nope, classic Arch Linux x64.

mdrv:
I use Arch Linux x64 as well but the bug only affects my arm64 PC somehow. That said, #588 might be able to resolve your issue (since it removes `-x` build option on both x64 and arm64 GNU targets).

PR #588, opened 2026-03-27
Fix Linux GNU aarch64 build by removing '-x' build option

mdrv:
## What is the motivation?

The Linux arm64 GNU binary shipped in `@surrealdb/node@3.0.3` (and below) fails to load on most Linux distributions (Arch Linux, Debian, Fedora, Ubuntu) due to a C++ ABI incompatibility introduced by Zig cross-compilation. You can check the test result [here](https://github.com/mdrv/surrealdb.js/actions/runs/23659970047).

```sh
bun add v1.3.11 (af24e281)
Resolving dependencies
Resolved, downloaded and extracted [28]
Saved lockfile

installed @surrealdb/node@3.0.3

396 | 	} catch (err) {
397 | 		if (process.env.NAPI_RS_FORCE_WASI) loadErrors.push(err);
398 | 	}
399 | }
400 | if (!nativeBinding) {
401 | 	if (loadErrors.length > 0) throw new Error("Cannot find native binding. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.", { cause: loadErrors });
                                            ^
error: Cannot find native binding. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
      at /tmp/test/node_modules/@surrealdb/node/dist/surrealdb-node.mjs:401:39
      at loadAndEvaluateModule (2:1)

Bun v1.3.11 (Linux arm64)
Error: Process completed with exit code 1.
```

## What does this change do?

Removes the `-x` flag from the `zigbuild` invocation for the Linux GNU targets in `publish-surreal.yml`. Without `-x`, Zig acts purely as a linker without injecting its own C++ runtime overrides, producing a binary that links correctly against the system `libstdc++` on any glibc-based distribution. The MUSL targets are unaffected and unchanged.

Tested via [GitHub Actions](https://github.com/mdrv/surrealdb.js/actions/runs/23659970047) covering 4 distros × 2 architectures (x86_64 + arm64) using the fixed package published as `@mdrv/surrealdb-node` vs. `@surrealdb/node`:

| Distro | x86_64 | arm64 |
|---|---|---|
| Ubuntu 24.04 | ✅ | ✅ |
| Debian bookworm | ✅ | ✅ |
| Fedora (latest) | ✅ | ✅ |
| Arch Linux | ✅ | ✅ |
| Alpine (musl) | ✅ | ✅ |

All arm64 runners fail with the upstream package and pass with the fixed build. You can verify independently by installing `@mdrv/surrealdb-node`.

- https://github.com/surrealdb/surrealdb.js/issues/575
- https://github.com/surrealdb/surrealdb.js/issues/585

### DIFF
--- a/.github/workflows/publish-surreal.yml
+++ b/.github/workflows/publish-surreal.yml
@@ -103,7 +103,7 @@ jobs:
           - name: Linux x86_64 (gnu)
             host: ubuntu-22.04
             target: x86_64-unknown-linux-gnu
-            build: bun run build:node --target x86_64-unknown-linux-gnu -x
+            build: bun run build:node --target x86_64-unknown-linux-gnu
 
           - name: Linux x86_64 (musl)
             host: ubuntu-22.04
@@ -118,7 +118,7 @@ jobs:
           - name: Linux aarch64 (gnu)
             host: ubuntu-22.04
             target: aarch64-unknown-linux-gnu
-            build: bun run build:node --target aarch64-unknown-linux-gnu -x
+            build: bun run build:node --target aarch64-unknown-linux-gnu
 
           - name: Linux aarch64 (musl)
             host: ubuntu-22.04
--- a/.github/workflows/publish-surreal.yml
+++ b/.github/workflows/publish-surreal.yml
@@ -103,7 +103,7 @@ jobs:
           - name: Linux x86_64 (gnu)
             host: ubuntu-latest
             target: x86_64-unknown-linux-gnu
-            build: bun run build:node --target x86_64-unknown-linux-gnu -x
+            build: bun run build:node --target x86_64-unknown-linux-gnu
 
           - name: Linux x86_64 (musl)
             host: ubuntu-latest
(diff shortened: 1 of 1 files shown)


mdrv:
A few references explaining why `-x` is unneeded for GNU targets.

> **1. napi-rs official cross-build docs**
> https://napi.rs/docs/cross-build
>
> > "install `cargo-zigbuild` if you want to build for **Linux musl targets**"
>
> GNU targets are not mentioned. No `-x` is shown anywhere in the napi-rs recommended workflow for GNU.

---

> **2. Zig's own `build.zig` source**
> https://github.com/ziglang/zig/blob/master/build.zig
>
> > `use_zig_libcxx` — "If libc++ is needed, use zig's bundled version, **don't try to integrate with the system**"
>
> This is the smoking gun: `-x` explicitly opts out of system C++ runtime integration — wrong for GNU targets that already have `libstdc++`.

---

> **3. cargo-zigbuild README**
> https://github.com/rust-cross/cargo-zigbuild
>
> > "By default `--target` for `*-gnu` will have Zig implicitly build for a default version of glibc"
>
> Glibc targets work out of the box — `-x` is an additional override that actively breaks the system integration that works correctly by default.

mdrv:
@macjuul The native binding error still persists on upstream (tested locally). May I ask to merge this PR?

Issue #603, opened 2026-05-06
Bug: Multiple library versions or instances conflict fatally


knackstedt:
### Describe the bug

If you have a multi-repo setup or dependencies with hard version requirements, instanced objects do not work across different library instances due to usage of `instanceof`. Additionally, private fields in the exported classes cause type conflicts because the classes are exported as-is, rather than exported as interface types.

### Steps to reproduce

Setup a bun multi-repo workspace with surrealdb installed in multiple repos. Make one workspace contain types and utilities, and have at least one library that either depends on or is a peer dependency of Surrealdb. You'll observe that bun creates multiple bun installs for the different contexts, and instanceof checks (e.g. with RecordIds and other instanced classes) will not resolve correctly.

e.g.
Instanceof checks don't correctly resolve RecordIDs -- they emit `{ }` to the underlying query.
```
Cannot execute DELETE statement using value: {  }","status":500,"stack":"Error: Cannot execute DELETE statement using value: {  }
```
Type errors because of #private properties not being rectifiable.
```
Types of parameters 'db' and 'db' are incompatible.
    Type 'import("/home/knackstedt/TEST_REPO/node_modules/.bun/surrealdb@2.0.3+4b2c1f8f02d6f8a0/node_modules/surrealdb/dist/surrealdb").Surreal' is not assignable to type 'import("/home/knackstedt/TEST_REPO/node_modules/.bun/surrealdb@2.0.3+85580646e632dbd6/node_modules/surrealdb/dist/surrealdb").Surreal'.
      Property '#private' in type 'Surreal' refers to a different member that cannot be accessed from within type 'Surreal'
```

### Expected behaviour

Passing an instanced RecordId from v2.0.0 to v2.0.3 should run correctly, without relying on an instanceof check. Additionally, types can resolve to different libraries, but should only conflict if there's an interface delta between the versions, not because the #private fields can't be rectified between the two.

### SurrealDB version

Irrelevant.

### JavaScript SDK version

2.x.x

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct (repository /surrealdb.js)

- [x] I agree to follow this project's Code of Conduct

PR #604, opened 2026-05-06
fix: handle multiple library instances (versions, duplicate dependencies)

knackstedt:

## What is the motivation?

When you have multiple SurrealDB SDK libraries imported (due to module resolution or version incompatibility) then if you pass e.g. a RecordId from library =2.0.0 to =2.0.1 the instanceof checks fail and the RecordId is serialized as `{ }` into the query.

e.g.
```
Cannot execute DELETE statement using value: {  }","status":500,"stack":"Error: Cannot execute DELETE statement using value: {  }
```

Additionally, there are type conflicts if multiple SurrealDB SDKs are resolved by the SDK -- specifically with Typescript #private fields. This can only be resolved by completely omitting the fields in the exported types or by marking them as normal private fields, without the #.

## What does this change do?
This change replaces all of the instanceof checks and #private fields with standard typescript private fields. 

Ensuring unit tests work. This is difficult to multi-instance test with the time I have available.

#603


macjuul:
Hey! I'm on board with making these changes, however one improvement that can be made to simplify and streamline the instance checking by implementing [Symbol.hasInstance](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/hasInstance) on each class, which in turn checks for symbol presence

knackstedt:
@macjuul I reverted those checks and moved everything to use Symbol.hasInstance to replace the util export checks.

Note: the Biome formatter and the VSCode formatters are conflicting in some syntax here -- might want to disable the VSCode formatter via a settings.json config property

macjuul (approved):

PR #617, opened 2026-06-02
Add protocol validation for WebSocket-only features (sessions, transa…

knackstedt:
…ctions, live queries)

## What is the motivation?

Throw clear errors when trying to access features unavailable with the current connection protocol

## What does this change do?

Throws errors when code tries to use SDK features available only via WS/WSS when connected over HTTP(S)

Tests didn't regress, this is a simple PR.

No

### DIFF
--- a/packages/sdk/src/api/session.ts
+++ b/packages/sdk/src/api/session.ts
@@ -1,4 +1,5 @@
 import type { ConnectionController } from "../controller";
+import { UnsupportedProtocolFeatureError } from "../errors";
 import type {
     AccessRecordAuth,
     AnyAuth,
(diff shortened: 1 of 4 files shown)


macjuul:
Looking through the changes, besides the doc comment improvements, I don't quite understand the goal achieved by this PR

In the current implementation we already throw a distinct `UnsupportedFeatureError` error when a given protocol doesn't support a specific feature, while `UnavailableFeatureError` is thrown when the remote version doesn't support a given feature.

Because of this the added checks in this PR should not make any difference to the way the SDK already operates, besides resulting in a different error class (which may also be considered a breaking change). Could you explain the reasoning behind this change further?

knackstedt:
I was under the impression that the SDK didn't throw errors, because in a few discrete states it simply wouldn't (though to be fair I haven't checked it in more recent versions). 
I'm fine with either closing this or stripping out the guards that I added

PR #627, opened 2026-06-18
ci: use ubuntu-latest and remove disabled build-wasm job

tobiemh:
## What changed

- Switched all `ubuntu-22.04` runners to `ubuntu-latest`:
  - `build-node` job in `.github/workflows/check.yml`
  - Four matrix hosts in `.github/workflows/publish-surreal.yml` (Linux x86_64/aarch64, gnu + musl)
- Removed the commented-out `build-wasm` job from `.github/workflows/check.yml`, along with its unused `runner-amd64-2xlarge` custom (self-hosted) runner reference.

## Why

The repo no longer uses any custom/self-hosted runners — the only reference (`runner-amd64-2xlarge`) was in a disabled, commented-out job, so it's dead weight. Standardizing on `ubuntu-latest` removes pinned OS versions so jobs track the current GitHub-hosted image.

## Reviewer notes

- All jobs now run exclusively on `ubuntu-latest`, `macos-latest`, or `windows-latest` (standard GitHub-hosted runners).
- The `ubuntu-22.04` pins were on the Linux native-module builds, which cross-compile via `cargo-zigbuild` — zig controls the glibc target independently of the runner OS, so bumping to 24.04 (what `ubuntu-latest` currently tracks) should not change the produced binaries. Worth a sanity check on the publish workflow if you want to be cautious.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -136,6 +136,65 @@ jobs:
           retention-days: 1
           path: packages/spectron/dist/
 
+  # build-wasm:
+  #   name: Build WASM
+  #   needs: build-sdk
+  #   runs-on: runner-amd64-2xlarge
+  #   steps:
+  #     - name: Install Bun
+  #       uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
+  #       with:
+  #         bun-version: 1.2.21
+
+  #     - name: Code Checkout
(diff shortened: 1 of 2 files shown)


kearfy:
Please do not remove the disabled build-wasm job. It should be added back due course, but is disabled right now I believe due to some issues with the library

tobiemh:
Restored the commented-out `build-wasm` job in 61a94df — kept it disabled as-is. The PR now only switches `ubuntu-22.04` → `ubuntu-latest`.

kearfy (approved):

macjuul (approved):

PR #641, opened 2026-07-08
Guard SharedArrayBuffer reference in Uuid constructor

### LINKED ISSUE #636: Bug: UUID creation throws on the client
### Describe the bug

Essentially every UUID constructor throws `ReferenceError: SharedArrayBuffer is not defined` on non-cross-origin-isolated pages

In @surrealdb/sqon (Uuid constructor), the line `if (uuid instanceof ArrayBuffer || uuid instanceof SharedArrayBuffer)` references SharedArrayBuffer unguarded. On any page that isn't cross-origin isolated, that global doesn't exist, and evaluating the instanceof right-hand operand throws a ReferenceError before any comparison happens. This essentially means that every client-side Uuid construction throws.

Fix is a one-token guard: typeof SharedArrayBuffer !== 'undefined' && uuid instanceof SharedArrayBuffer.

### Steps to reproduce

Try to create any UUID on the client on non-cross-origin-isolated pages.

### Expected behaviour

UUID creation works

### SurrealDB version

3.1.5 for macos on aarch64

### JavaScript SDK version

2.0.4

### Contact Details

_No response_

- [x] I have searched the existing issues

### Code of Conduct (repository /surrealdb.js)

- [x] I agree to follow this project's Code of Conduct

apoorva-01:
## What is the motivation?

`SharedArrayBuffer` isn't a defined global on pages that aren't cross-origin isolated. The `Uuid` constructor referenced it unguarded as its first check, so `uuid instanceof SharedArrayBuffer` threw `ReferenceError` before it could compare anything. Every client-side construction hit it, including `new Uuid("...")` and `Uuid.v4()`.

## What does this change do?

Guards the reference with `typeof SharedArrayBuffer !== "undefined"`, wrapped in a small `isSharedArrayBuffer` type-guard. Inlining the check defeats TypeScript's narrowing and breaks the typecheck at `UUID.parse`, so the helper keeps it clean and matches the `typeof` guards already used in `datetime.ts`/`duration.ts`.

Added a regression test that deletes `globalThis.SharedArrayBuffer`, constructs a Uuid, then restores the global. It throws without this change and passes with it. Full sqon unit suite stays green.

Closes #636

### DIFF
--- a/packages/sqon/src/value/uuid.ts
+++ b/packages/sqon/src/value/uuid.ts
@@ -3,6 +3,12 @@ import { JsonCodec } from "../codec/json/codec.ts";
 import { hasSymbol, markSymbol, UUID_SYMBOL } from "../utils/symbols.ts";
 import { Value } from "./value.ts";
 
+// SharedArrayBuffer is only exposed on cross-origin-isolated pages, so
+// referencing it directly throws a ReferenceError everywhere else.
+function isSharedArrayBuffer(value: unknown): value is SharedArrayBuffer {
+    return typeof SharedArrayBuffer !== "undefined" && value instanceof SharedArrayBuffer;
+}
+
 /**
  * A SurrealQL UUID value.
  */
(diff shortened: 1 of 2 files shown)


kearfy (approved):

Nerivec:
Nitpick: I'd inline instead of function decl/call.

Note that this breaks Vite dev server use for example.
Workaround:
```ts
    plugins: [
        {
            name: "configure-response-headers",
            configureServer: (server) => {
                server.middlewares.use((_req, res, next) => {
                    res.setHeader("Cross-Origin-Embedder-Policy", "require-corp");
                    res.setHeader("Cross-Origin-Opener-Policy", "same-origin");
                    next();
                });
            },
        },
        // rest...
    ],
```

PR #664, opened 2026-08-23
WebSocket query streaming for JavaScript

tobiemh:
## What is the motivation?

The server can now answer a query with a sequence of frames instead of one response ([surrealdb-private#812](https://github.com/surrealdb/surrealdb-private/pull/812), following #790's gRPC work), so the first rows can reach a client while the query is still running. That PR's own closing note is this change: the SDK can adopt it invisibly — send `query_stream`, accumulate frames, resolve the same promise, and fall back to `query` when the server does not know the method.

Users get a lower time to first result and no single enormous response to decode, with no API change. The visible `for await (const row of …)` API is a later, thin layer over frames the SDK now already parses.

## What does this change do?

The WebSocket engine asks for queries with `query_stream` and translates the frames into the `QueryChunk` stream that `collect()`, `responses()` and `.stream()` already consume. Nothing above the engine changed.

**Adoption is invisible.** The streamed answer is the buffered answer — same chunks, same errors, same statistics. Three things make that true rather than hoped for:

- A server which does not serve the method answers `-32601`, and the query is asked for again in buffered form. Same for a denial (`-32602`), for a stream refused because the connection is at its concurrency cap, and for one whose socket went away before anything was framed. Nothing is framed until execution is about to begin, so in every one of those cases the query never ran and re-asking cannot execute it twice. Absent-or-denied is remembered for the life of the socket; a capacity refusal is not, since it is transient.
- Queries inside a transaction from `.begin()` are never streamed: a `commit` can arrive on the same connection mid-stream, and would commit a prefix of the query rather than the whole of it.
- `DriverOptions.streaming` turns it off entirely.

**The frame contract is preserved as it is translated.** Rows are provisional until their statement's terminal frame; a statement's failure becomes the error chunk that retracts the rows already delivered for it; a stream stopped rather than answered fails the query, so a truncated answer is never mistaken for a whole one; results are counted by terminal frames, never by the statement count on `begin`, which is only an upper bound. Abandoning a stream — breaking out of a `for await` — cancels it on the server instead of leaving it to produce results nothing will read.

### How is it used?

Nothing changes for a caller who wants the whole answer. The query below is streamed on a server
which serves it and buffered on one which does not, and returns the same 300 records either way:

```ts
const [people] = await db.query<[Person[]]>("SELECT * FROM person").collect();
```

`.stream()` is what the streaming is for: the rows arrive as the server produces them, so a
statement's first row does not wait on the rest of the query. Each frame belongs to a statement,
by index, and `isValueOf` / `isDoneOf` check both the kind of frame and which statement it is for:

```ts
for await (const frame of db.query("SELECT * FROM person; SLEEP 500ms;").stream()) {
    if (frame.isValueOf<Person>(0)) {
        console.log(frame.value.age);       // a row of the first statement
    }

    if (frame.isDoneOf(0)) {
        console.log("that statement is complete");
    }

    if (frame.isErrorOf(0)) {
        // The statement failed. Every row above for statement 0 is retracted.
    }
}
```

Against a server with #812 that prints the first row **11ms** in, while the query itself completes
at **512ms** - the `SLEEP` is the rest of the query, and the rows no longer wait behind it.

Note that the `<Person>` on `isValueOf` asserts the row type rather than checking it: `query()`
takes a tuple of statement types and the frame API does not derive from it, so it is on the caller
to keep the two in agreement. Deriving it, and a row-level `for await (const person of ...)` view
over these frames, are both worth having and are deliberately not in this change - see the
discussion on this PR.

The builders need no such annotation, because each is a single statement and its result type says
what its frames carry:

```ts
for await (const frame of db.select<Person>(person).stream()) {
    if (frame.isValue()) console.log(frame.value.age);   // Person, inferred
}
```

That was typed by the statement's result rather than the row until this change, so a frame of a
`select` over a table declared `Person[]` while carrying one `Person` - `frame.value.age` did not
compile and `frame.value[0]` did. `select`, `create`, `update`, `upsert`, `delete`, `insert` and
`relate` now read the row off their result type; `run`, `auth` and `api` are single-value
statements and keep carrying their result whole. The public API typecheck fixture pins all four
shapes.

Leaving the stream stops the query rather than letting the server produce results nothing will
read, and the connection is immediately usable again:

```ts
for await (const frame of db.query("SELECT * FROM person; SLEEP 30s;").stream()) {
    if (frame.isValue<Person>()) break;      // sends `query_cancel` for this query
}

await db.query("RETURN 1").collect();        // answers straight away
```

And it can be turned off for the whole driver, which puts every query back on the buffered path:

```ts
const db = new Surreal({ streaming: false });
```

### Two fixes this depends on

- **`ChannelIterator` lost values.** A value submitted before the consumer asked for the next one resolved the same, already settled promise, so every value but the first of a burst vanished. This was dropping **live query notifications** too, not only frames. Readers are now queued, which also settles a reader that a second reader used to abandon forever.
- **`collect()` and `responses()` rewrote the chunks they were built from.** They adopted the array a chunk carried and pushed later batches into it. Harmless while every statement arrived as one chunk; with batches, a diagnostics subscriber holding a chunk that carried 16 rows would later find 499 in it.

### One deliberate non-change

A frame carries the full error object, wire code included, where a buffered query result carries none and `ServerError.code` is `0`. The code a frame carries is dropped, so the same failing query does not report a different code depending on whether its answer happened to be streamed. `QueryChunk.type` is likewise left unset for an ordinary statement, exactly as the buffered path leaves it — the consumers of a chunk are what settle on a default.

Verified against a build of the server **with** #812 and, separately, against one **without** it, so both the streaming path and the fallback are covered by the whole suite.

| Gate | Result |
|---|---|
| `bun test` (WebSocket, server **with** #812) | 618 pass, 0 fail |
| `bun test` (WebSocket, server **without** #812 — fallback) | 618 pass, 0 fail |
| `SURREAL_PROTOCOL=http bun test` (server with #812) | 583 pass, 0 fail |
| `bun run qc`, `bun run qts`, `bun run build:sdk` | clean |
| `tsc --project packages/tests/surrealdb/typecheck` (public API, against the built `.d.ts`) | clean |

Reproduce the streaming run with `SURREAL_EXECUTABLE_PATH=<surreal with #812> bun test --preload ./global.ts` from `packages/tests/surrealdb`. On a debug build add `--timeout 30000`: `sessions > restore state after reconnect` restarts the database inside bun's 5s default and cascades when it cannot.

**New tests**, chosen so they cannot pass for the wrong reason:

- `unit/streaming/frames.test.ts` — the translator against a scripted sequence of frames: batching and per-statement batch numbering, a single value held back until its statement says it was one, `NONE`, statement failure, stream failure retracting only the statements that never finished, a truncated stream, terminal-once, `begin.statements` as an upper bound, an unrepresentable index, an unreadable duration, and the error-code and type parity described above.
- `unit/streaming/results.test.ts` — the same scripts through `collect()`, `responses()` and `.stream()`, which is where retraction has to be observable: an errored statement drops the rows it delivered while its neighbours stand.
- `unit/streaming/engine.test.ts` — the engine over a mock `WebSocket` speaking the real CBOR protocol, so frame ordering, correlation by request id and cancellation are exercised for real: the request shape, a 2000-row burst arriving complete and in order, two concurrent streams keeping their own frames, cancel on abandonment and no cancel after a clean end, both fallback codes, a capacity refusal falling back for one query only, a non-frame payload, a stream never replayed once it has framed anything, and both reconnect paths.
- `integration/query/streaming.test.ts` — against a real server, including a differential test asserting a streamed answer equals the same query on a `streaming: false` connection, and one asserting that the failures of a mixed query match across the two paths field for field. A test which needs to know streaming actually happened asks a raw socket rather than the SDK, whose whole point is to hide the difference — so it hard-fails if streaming ever silently degrades to the fallback on a capable server.
- `unit/utilities/channel-iterator.test.ts` — pins the lost-value fix directly rather than by the mass timeout it used to cause.

The usage above was run against the #812 build as written, and the timings quoted in it are from that run. Beyond the suite, the streaming path was driven end to end against the same build: frames arrive as `begin{2} → rows(16) → rows(23) → finished → value → finished{single} → end`; a 499 row result reaches the SDK as chunks of 16, 32, 64, 128, 256, 3; the first row lands 5.9ms in against a query that completes at 1008ms; abandoning a `SLEEP 30s` stream returns in 6.7ms with the connection immediately reusable; and a `LIVE SELECT` registered through a stream delivers its notifications.

Implements the JS SDK side of surrealdb-private#812 (WebSocket streaming), which follows surrealdb-private#790 (gRPC end-to-end streaming).

Not in scope here: the `for await (const row of …)` API this makes possible, and a `Features.QueryStreaming` version gate — with the release that ships #812 known, the SDK could skip the one probe it currently spends per connection against older servers.

### DIFF
--- a/packages/sdk/src/engine/websocket.ts
+++ b/packages/sdk/src/engine/websocket.ts
@@ -9,6 +9,7 @@ import {
 } from "../errors";
 import { parseRpcError } from "../internal/parse-error";
 import {
+    type Abandonment,
     isQueryStreamFrame,
     type QueryStreamFrame,
     queryStreamChunks,
@@ -189,6 +190,10 @@ export class WebSocketEngine extends RpcEngine implements SurrealEngine {
         this.#terminated = true;
         this.#socket?.close();
 
+        // No socket is coming now, so the streams waiting for one are settled here rather than
+        // when the reconnect cooldown they are waiting through happens to elapse.
+        this.failStreams(new CallTerminatedError());
+
         if (socketState === WebSocketImpl.OPEN || socketState === WebSocketImpl.CLOSING) {
             await this.#publisher.subscribeFirst("disconnected");
         } else {
@@ -240,7 +245,7 @@ export class WebSocketEngine extends RpcEngine implements SurrealEngine {
         });
     }
 
-    override async *query<T>(
+    override query<T>(
         query: BoundQuery,
         session: Session,
         txn?: Uuid,
@@ -258,16 +263,52 @@ export class WebSocketEngine extends RpcEngine implements SurrealEngine {
             this._context.options.streaming === false ||
             !this.#socket
         ) {
-            yield* super.query<T>(query, session, txn);
-            return;
+            return super.query<T>(query, session, txn);
         }
 
+        // The frames are kept to hand so that leaving the chunks can act on them at once. An
+        // async generator defers a `return` until its body next wakes, and this body can be
+        // parked on a frame which is not coming - a consumer racing a read against a timeout
+        // would otherwise never be let go.
+        const frames: { current?: AsyncIterableIterator<QueryStreamFrame> } = {};
+        const abandonment: Abandonment = {};
+        const chunks = this.streamChunks<T>(query, session, txn, frames, abandonment);
+
+        return {
+            [Symbol.asyncIterator]: () => ({
+                next: () => chunks.next(),
+                throw: (error?: unknown) => chunks.throw(error),
+                return: (value?: QueryChunk<T>) => {
+                    // Recorded before the frames are closed, so that running out of them is read
+                    // as the consumer leaving rather than as a truncated answer.
+                    abandonment.requested = true;
+                    frames.current?.return?.(undefined);
+                    return chunks.return(value as QueryChunk<T>);
+                },
+                [Symbol.asyncIterator]() {
+                    return this;
+                },
+            }),
+        };
+    }
+
+    /**
+     * Streams a query, falling back to the buffered method if the server refuses it.
+     */
+    private async *streamChunks<T>(
+        query: BoundQuery,
+        session: Session,
+        txn: Uuid | undefined,
+        frames: { current?: AsyncIterableIterator<QueryStreamFrame> },
+        abandonment: Abandonment,
+    ): AsyncGenerator<QueryChunk<T>> {
         const probe: StreamProbe = { framed: false };
 
+        // Opened here rather than by `query`, so nothing is sent until the caller reads.
+        frames.current = this.streamFrames(query, session, probe);
+
         try {
-            for await (const chunk of queryStreamChunks<T>(
-                this.streamFrames(query, session, probe),
-            )) {
+            for await (const chunk of queryStreamChunks<T>(frames.current, abandonment)) {
                 yield chunk;
             }
         } catch (error) {
@@ -285,7 +326,9 @@ export class WebSocketEngine extends RpcEngine implements SurrealEngine {
                 this.#streaming = false;
             }
 
-            yield* super.query<T>(query, session, txn);
+            for await (const chunk of super.query<T>(query, session, txn)) {
+                yield chunk;
+            }
         }
     }
 
@@ -641,7 +684,7 @@ export class WebSocketEngine extends RpcEngine implements SurrealEngine {
  * re-sent, because the second request would have to be started from a stream whose consumer may
  * be walking away at that very moment, leaving its rejection with nowhere to go.
  */
-function isRetriableAsBuffered(error: unknown): boolean {
+function isRetriableAsBuffered(error: unknown): error is ServerError {
     return error instanceof ServerError;
 }
 
--- a/packages/sdk/src/engine/websocket.ts
+++ b/packages/sdk/src/engine/websocket.ts
@@ -45,10 +45,14 @@ interface Call<T> {
  */
 interface Stream {
     channel: ChannelIterator<StreamEvent>;
+    /** The request which opened it, re-sent if a socket returns before it framed anything. */
+    request: object;
     /** Whether the terminal frame, or a failure standing in for it, has arrived. */
     settled: boolean;
     /** Whether the consumer stopped reading, leaving remaining frames to discard. */
     abandoned: boolean;
+    /** Whether any frame has been delivered, which is what makes a replay unsafe. */
+    probe: StreamProbe;
 }
 
 /**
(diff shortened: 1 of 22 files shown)


itsezc:
My two cents for DX:

1. `isValueOf<Person>(0)` is an unchecked cast wearing a type-guard costume. The buffered path already carries the tuple type: `db.query<[Person[]]>(...)`. The streaming path throws it away and asks you to re-declare `<Person>` per call site, where nothing validates it against the query type. That's two places to keep in sync and one of them lies. It's derivable:

```ts
const q = db.query<[Person[], Company[]]>("SELECT * FROM person; SELECT * FROM company");
for await (const frame of q.stream()) {
  if (frame.isValueOf(0)) frame.value.age;   // Person, inferred
  if (frame.isValueOf(1)) frame.value.name;  // Company, inferred
}
```

An indexed-access conditional `(T[I] extends (infer R)[] ? R : T[I])` gets you there, and `isValueOf(7)` on a 2-statement query becomes a compile error instead of a branch that silently never fires.

2. The bigger DX win: don't make people filter by index at all. Frame-level iteration is the low-level primitive; almost every real use is "give me the rows of statement N as they land." A per-statement view collapses the whole block:

```ts
// single statement — the 90% case
for await (const person of db.query<[Person[]]>("SELECT * FROM person").rows()) {
  console.log(person.age);
}

// multi-statement
const [people, companies] = db.query<[Person[], Company[]]>(q).statements();
for await (const p of people) { ... }
```

No guards, no magic index, no isDoneOf — the loop ending is done, and errors throw. Keep .stream() underneath for people who genuinely want frames.

tobiemh:
Both criticisms are correct, and I've not taken either into this PR. Reasoning, so the call is visible rather than implied:

**On the unchecked cast.** You're right that it lies: `isValueOf<V = T>(query: number)` asserts `V` and accepts any `number`, so `isValueOf<Company>(0)` on a `[Person[]]` query compiles, and `isValueOf(7)` on a two-statement query is a branch that can never fire. The tuple is right there on `Query<R>` and `stream()` discards it.

What makes it a separate change rather than a small one here: inference needs the index in the first type-parameter position — `isValueOf<I extends StatementIndex<R>>(query: I): this is ValueFrame<StatementRow<R, I>, J>` — which breaks every existing `isValueOf<Person>(0)`, i.e. exactly the call shape the current docs recommend. `Frame`, `ValueFrame`, `ErrorFrame` and `DoneFrame` are all exported from the package root, so their generic arity is public too. That's a deliberate breaking typing change, and this PR's premise is that adopting streaming changes nothing observable — mixing the two would make it impossible to adopt one without the other.

**On the per-statement view.** Agreed, and this is the better of the two. #812's own closing note says the same thing — that the visible `for await (const row of …)` API is "a later, thin layer over frames the SDK would already parse" — which is what this PR builds rather than exposes. `rows()` for the single-statement case is additive and non-breaking and could land on its own.

`statements()` needs one design decision first: frames arrive in execution order, so with N independent iterables a consumer that reads statement 1 before statement 0 either stalls or forces us to buffer statement 0 in full. So the API has to either promise sequential consumption or state that it buffers — worth settling deliberately rather than in this PR.

**Changed here, from your first point:** the usage section no longer presents `isValueOf<Person>(0)` as type-safe — it says the type argument is asserted, not checked, and points at this discussion. Separately, `ba900bd` fixed the `stream()` docs, which claimed an index could be passed to `isValue`/`isError`/`isDone`; those take none and ignore one, so following the docs silently matched every statement's frames. My own first draft of the example hit exactly that.

